PR #267, opened 2020-03-16
Chevron are correct size after removal of .site-wrapper class

emmabckstrm:
This is not code I am proud of, but a quick fix that will be discared in the future as we migrate away from custom css

### DIFF
--- a/app/assets/stylesheets/base.scss
+++ b/app/assets/stylesheets/base.scss
@@ -169,22 +169,43 @@ div.legal {
 }
 
 /* Glyphicon chevron overrides */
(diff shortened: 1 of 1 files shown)


stefanlindbohm:
Maybe just drop the `.site-wrapper` prefix? Or would that affect glyphicon usage on new clean pages, no?

emmabckstrm:
@stefanlindbohm that was my first thought but it didn't work as the icons also has the class `glyphicon-large` (which gives the greenish color and size) and that rule won

stefanlindbohm:
> @stefanlindbohm that was my first thought but it didn't work as the icons also has the class `glyphicon-large` (which gives the greenish color and size) and that rule won

Ugh. OK, let's go.

PR #268, opened 2020-03-17
First iteration of breakit travel calculator

kallenilver:


kallenilver:
@stefanlindbohm @emmabckstrm I didn't spend time on learning where to put the js or css correctly, and not how to adapt to the new tailwind structure. But I think the biggest risk here is that I've  built something that breakit doesn't want. So my plan is to deploy a stage-app to heroku and get feedback before writing something more maintainable  :)

config/locales/sv.yml
```diff
@@ -644,6 +644,22 @@ sv:
...
+        travel_emissions_calculator_description: Vill du räkna ut klimatavtrycket för en resa? Skriv in hur långt du ska åka här nedan så fixar vi resten!
+        distance_in_kilometers: Längd på resa i kilometer
+        calculate: Beräkna
+        emissions_if_traveling_by: Klimatutsläpp om du reser med
+        flight: Flyg
+        car-petrol: Bil (bensin)
+        car-diesel: Bil (diesel)
+        car-etanol: Bil (etanol)
+        car-gas: Bil (gas)
+        car-electric: Elbil
```
emmabckstrm: should the keys be using underscore instead of dash?

emmabckstrm:
@kallenilver , don't worry about using tailwind, it's not ready to be rolled out in our app yet, we have some decisions to take before rolling out, and we're prioritizing the calculator2.0 higher :) there will be some sort of communication when it's time to use tailwind!

app/views/welcome/travel_calculator.html.erb
```diff
@@ -0,0 +1,103 @@
...
+          var distance = parseInt(document.getElementById("distance").value);
+
+          var emissionTypes = [["flight", 0.25], 
+                               ["car-petrol", 0.19],
+                               ["car-diesel", 0.15],
+                               ["car-etanol", 0.11],
+                               ["car-gas", 0.04],
+                               ["car-electric", 0.01],
+                               ["bus",  0.059],
+                               ["train",  0],
```
stefanlindbohm: This isn't really accurate, is it? Train (and subway) should have some footprint, even though it's small. Are these final numbers?

config/locales/sv.yml
stefanlindbohm: Agree

stefanlindbohm:
@kallenilver This is getting stale. Merge or kill? :)

kallenilver:
> @kallenilver This is getting stale. Merge or kill? :)

There might be some value to it, so if the code quality is ok with you I'm voting for merge.

app/views/welcome/travel_calculator.html.erb
```diff
@@ -0,0 +1,103 @@
...
+                               ["subway",  0],
+                               ["boat",  0.17],
+                              ];
+
+          emissionTypes.map(function(emissionType) {
+            var stapleLength = emissionType[1] * 250 / 0.25;
```
stefanlindbohm, line 76: I think you mean bar length? :)

stefanlindbohm:
> > @kallenilver This is getting stale. Merge or kill? :)
> 
> There might be some value to it, so if the code quality is ok with you I'm voting for merge.

Sure. But please adress the comments before merging 👍

app/views/welcome/travel_calculator.html.erb
kallenilver: The train emissions are so small (if in Sweden) it's very close to 0.

PR #269, opened 2020-03-18
Create section partial (with tailwind)

emmabckstrm:


stefanlindbohm:
Looking at this in practice, I must say I'm a little hesitant for this. These sections here are really very specific variants of sections where we have only text. If we do this, maybe it should be `text_section`. But then again, sooner or later we'll run into edge cases where we want to tweak something in a specific way. Currently leaning towards it being better to keep the duplication–after all it's simple things repeated and we can always search and replace if we need to change them all. Where's your thinking on this right now?

emmabckstrm:
@stefanlindbohm My take is that I've seen a pattern in our current design that we use this same section in multiple places all over the page. I think it's a good idea to use a cohesive design, if the sections look similar with different content. 
I do however also think it's nice to keep the code DRY, and I really see that we could leverage a partial like this when we transition into the new design, then we only need to update the partial(s), instead of going into every file and change every section manually.

I do agree with you on the section could be called something more specific, or at least we could try to figure out a better system of deciding on how specific the partial should be. What I wrote now is adapted to what we have today, and of course I think we could either develop this partial to be more flexible, for instance having locals for images, and options for left or right alignment etc, OR if we divide it up into multiple partials, for instance `section_text`, `section_cta`, `section_image`. 

Edge cases will always be edge cases, and I think it's pretty much impossible to expect a partial like this to cover every possible use case, but I do think it's possible to create partials like this that cover the vast majority of the use cases.

stefanlindbohm:
@emmabckstrm Now that this is getting stale, I'd vote we close this PR and do something like this when we start the bigger job of implementing our new design. What say you?

stefanlindbohm:
Will get revived in some way in #284

PR #271, opened 2020-03-24
New calculator on a hidden URL

stefanlindbohm:
## Todo

- [x] Why is the green circle not working in Safari on staging? -- It was being purged by PurgeCSS because we moved JS files out of where PurgeCSS was looking
- [x] Fix validations in Safari (radio buttons are marked as not checked?) -- Radio buttons were not checked because we had prevented default event behavior
- [x] Fix failing tests/saving of calculator for footprints
- [x] Add missing tests for `LifestyleFootprint`
- [x] Ensure things don't crash when the user isn't filling in flight hours (and maybe car distance?)
- [x] Refactor `LifestyleFootprintController` (.js)
- [x] Make highlight styling nicer for inputs
- [x] Fix going back when in the first category
- [x] Review what URL the calculator should be available on
- [x] Use keys to reference footprint records in URL's instead of ID's that are easily tinkered with.
- [x] Multiply car distance by 52 ...somewhere? (Since we're asking for a distance per week)
- [x] Fix missing styling for flight hours field on iPhone Safari
- [x] Something is triggering side scrolling on iPhone Safari after focusing the input fields, but not on our other pages. What is happening?--text size was slightly under 16px, so the page was zoomed in just a little
- [x] Add localization
- [x] Add Swedish & <s>German</s> translations
- [x] Squash commits


stefanlindbohm:
History for step-by-step review: https://github.com/goclimateneutral/goclimateneutral/tree/calculator-page-history

config/locales/en.yml
```diff
@@ -571,6 +571,72 @@ en:
...
+          geothermal: Geothermal pump
+          heating_oil: Heating oil
+          propane: Propane
+          dont_know: I don't know/Other
+        green_electricity:
+          'yes': Yes
+          seventyfive_percent: 75% green
+          fifty_percent: 50% green
+          twentyfive_percent: 25% green
+          'no': No
```
emmabckstrm: why are yes and no quoted here?

stefanlindbohm: Because apparently `yes` and `no` are magical in Rails internationalization and get turned into `true`/`false` 😬 . Quoting leaves them as is.

stefanlindbohm: Oh wait. They're keywords in YML. That's why.

PR #272, opened 2020-04-03
Update admin fee and price for climate report

kallenilver:


stefanlindbohm:
@kallenilver Are we doing this?

PR #273, opened 2020-04-08
Launch calculator 🚀

stefanlindbohm:


stefanlindbohm:
Maybe we should add that integration test now that we're taking this live? Also... maybe our current integration test will fail 5% of the time because we're not making sure it uses the previous/current version? Haha.

stefanlindbohm:
Maybe use country code reported by CloudFlare as priority country instead? That should be correct in 99%+ of cases.

PR #274, opened 2020-04-13
Disallow negative distances/flight hours in calculator

stefanlindbohm:
Before this change, you could give yourself a "discount" on your
footprint by entering negative numbers :).

Setting min="0" on <input type="number"> seems to be enough in both
Safari/WebKit and Firefox to also stop the stepper from going below
zero. Entering a negative value will still result in not being able to
click Next to finish the calculator, but hopefully this case is
extremely rare.

### DIFF
--- a/app/models/lifestyle_footprint.rb
+++ b/app/models/lifestyle_footprint.rb
@@ -13,8 +13,8 @@ class LifestyleFootprint < ApplicationRecord
   attribute :total, :greenhouse_gases
 
   validates :key, uniqueness: true, format: { with: /\A[a-f0-9]{40}\z/ }
-  validates_presence_of :housing, :food, :car, :flights, :consumption, :public, :total, :car_distance_answer,
-                        :flight_hours_answer
+  validates_presence_of :housing, :food, :car, :flights, :consumption, :public, :total
+  validates :car_distance_answer, :flight_hours_answer, numericality: { greater_than_or_equal_to: 0 }
   validate :answers_exist_in_calculator
 
   before_validation :generate_key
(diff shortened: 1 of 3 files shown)


emmabckstrm:
Great work! When doing like this, the user won't get any feedback on why they cannot click "Next" if they have a negative number, correct? For me this solution is not optimal due to not giving feedback to user, and they could think that the calculator is broken and just drop off. While if they get a negative amount of pay, that's still some sort of feedback, even though it's bad. An optimal solution for me would be to add a little warning/informaiton text "You can't travel a negative distance unless you're traveling in time." (:wink:).

stefanlindbohm:
> **Edit**: Okay, seems like once again I'm unaware of the rails magic, so nevermind my comment below! It gives feedback if I have a negative number in flight (last step) but not if I have a negative number in second to last step. Good enough!
> 
> Great work! When doing like this, the user won't get any feedback on why they cannot click "Next" if they have a negative number, correct? For me this solution is not optimal due to not giving feedback to user, and they could think that the calculator is broken and just drop off. While if they get a negative amount of pay, that's still some sort of feedback, even though it's bad. An optimal solution for me would be to add a little warning/informaiton text "You can't travel a negative distance unless you're traveling in time." (😉).

UX is actually affected only by client side validations (no Rails integration). Nothing happens because validations fail (`min` attribute on `<input type="text">` creates a validation). But nothing is shown for hidden input fields because they're hidden.

I added the Rails validations as a fall-back for if someone would fiddle with this and as a help for us in the future. And for our own sanity so we're absolutely certain we don't end up with nonsensical data in the DB.

PR #275, opened 2020-04-13
Persist country for LifestyleFootprint for future use

stefanlindbohm:
We'd like to be able to prompt users when updated calculators for their
country are published.

## Todo

- [x] Split into two PR's with the migration separate.

### DIFF
--- a/app/controllers/lifestyle_footprints_controller.rb
+++ b/app/controllers/lifestyle_footprints_controller.rb
@@ -23,7 +23,7 @@ class LifestyleFootprintsController < ApplicationController
 
   def footprint_params
     params.require(:lifestyle_footprint).permit(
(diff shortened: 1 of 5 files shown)


db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_03_19_174526) do
+ActiveRecord::Schema.define(version: 2020_04_13_073358) do
```
emmabckstrm, line 13: should this be removed from this PR?

stefanlindbohm: This change is in a commit that went into master in #276 but was also on this branch :).

PR #277, opened 2020-04-17
Generate business climate report with graphs

kallenilver:


stefanlindbohm:
@kallenilver Instead of putting "WIP" in the title, GitHub now has "draft" PR's. I switched this one to draft and removed the WIP title :).

kallenilver:
> @kallenilver Instead of putting "WIP" in the title, GitHub now has "draft" PR's. I switched this one to draft and removed the WIP title :).

Great, thanks!

kallenilver:
@stefanlindbohm @emmabckstrm We are not completely done with the new business report, but I think it's time for a code check-in, so if anyone of you want to take a look at the code and get back with feedback on the direction that would be great! Thx!

app/pdfs/climate_report_pdf.rb
```diff
@@ -0,0 +1,199 @@
...
+
+  def field_percentage(field)
+    BigDecimal(@cr.calculation.send("#{field[:name]}_emissions")) / @cr.calculation.total_emissions * 100
+  end
+
+  def pie_data(fields)
+    data = {}
+    fields.map do |field|
+      data[field_name(field)] = field[:percentage] unless field[:percentage] == 0
+    end
```
stefanlindbohm, line 103: This is using `map` as if it were `each`. Same result:

```ruby
data = fields.map do |field|
  [field_name(field), field[:percentage]] unless field[:percentage] == 0
end.compact.to_h
```

app/pdfs/climate_report_pdf.rb
```diff
@@ -0,0 +1,199 @@
...
+    { labels: "'#{data.keys.join("', '")}'", data: data.values.join(', ') }
+  end
+
+  def pie_scope_data
+    scopes = { 'Scope 2' => 0, 'Scope 3' => 0 }
+    emissions.map do |field|
+      scopes["Scope #{field[:scope][0]}"] += field[:emissions]
```
stefanlindbohm, line 110: Is it intentional to only pick the first scope if a line item is has multiple scopes? If so I think `.first` instead of `[0]` better communicates intent.

app/pdfs/climate_report_pdf.rb
```diff
@@ -0,0 +1,199 @@
...
+    data = {}
+    fields.map do |field|
+      data[field_name(field)] = field[:percentage] unless field[:percentage] == 0
+    end
+    { labels: "'#{data.keys.join("', '")}'", data: data.values.join(', ') }
+  end
+
+  def pie_scope_data
+    scopes = { 'Scope 2' => 0, 'Scope 3' => 0 }
+    emissions.map do |field|
```
stefanlindbohm, line 109: This should be `.each`, not `.map`.

app/pdfs/climate_report_pdf.rb
```diff
@@ -0,0 +1,199 @@
...
+      scopes["Scope #{field[:scope][0]}"] += field[:emissions]
+    end
+    scopes.each do |k, v|
+      scopes[k] = BigDecimal(v) / @cr.calculation.total_emissions * 100
+    end
+    data = get_even_percentages(scopes)
+    data.filter! { |_, v| v != 0 }
```
stefanlindbohm, line 116: I'm feeling this part on separating between scopes should be part of `ClimateReportCalculation`, as well as percentages calculation. So it itself knows about scopes in one place and if we poke around can use the same without having to dig in PDF generation logic to re-use it. Maybe have `scope_2_emissions`, `scope_3_emissions`, `scope_2_percentage` and `scope_3_percentage` as attributes on the model?

app/pdfs/climate_report_pdf.rb
```diff
@@ -0,0 +1,199 @@
...
+
+    bar_fields.each do |field|
+      next if field[:emissions] == 0
+
+      total_emissions = climate_reports.inject(0) do |n, cri|
+        n + cri.climate_report.calculation.send("#{field[:name]}_emissions")
+      end
+      data << { field_name(field) => field[:emissions].round / @cr.employees }
+      data << { field_average_name(field) => (total_emissions / total_employees).round }
+    end
```
stefanlindbohm, line 143: This is loading every single row from the DB as a record, right? I think this should be replaced with pure SQL using `SELECT SUM()`'s. That also removes the need for Ruby code to iterate and sum the different fields :).

app/pdfs/climate_report_pdf.rb
```diff
@@ -0,0 +1,199 @@
...
+    {
+      data: data.map { |d| d.values.join('') },
+      labels: data.map { |d| d.keys.join('') }
+    }
+  end
+
+  def climate_periods_to_compare
+    ClimateReport.joins(:invoice).where(company_name: @cr.company_name).count
```
stefanlindbohm, line 151: Why join `invoice`? It just counts so it's not used, right?

app/pdfs/climate_report_pdf.rb
```diff
@@ -0,0 +1,199 @@
...
+  end
+
+  def bar_compare_years_data(bar_fields, per_employee = false) # rubocop:disable Metrics/AbcSize
+    data = []
+    climate_reports = ClimateReport.joins(:invoice).where(company_name: @cr.company_name)
```
stefanlindbohm, line 156: Joining `invoice` seems unecessary here, but `calculation`, on the other hand, is used, so maybe join that?

Also, since it's used both for counting and for data, maybe do an accessor that stores its value for previous reports?

spec/factories/climate_report.rb
```diff
@@ -9,6 +9,11 @@
     calculation_period { '2018' }
     calculation_period_length { 'year' }
 
+    trait :old do
```
stefanlindbohm, line 12: Nicely used :+1:

app/controllers/admin/climate_reports_controller.rb
```diff
@@ -1,24 +0,0 @@
-# frozen_string_literal: true
-
-module Admin
-  class ClimateReportsController < AdminController
```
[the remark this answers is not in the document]
kallenilver: Yes, this was the csv-export thing now no longer needed. Good catch, I forgot about the route. Removing it.

app/controllers/business/climate_reports_controller.rb
```diff
@@ -4,9 +4,18 @@ module Business
...
-      @invoice = ClimateReportInvoice.new(climate_report: @report)
-                                     .tap(&:calculate_from_report)
-      @projects = Project.order(id: :desc).limit(2)
+      respond_to do |format|
+        format.html do
+          @invoice = ClimateReportInvoice.new(climate_report: @report).tap(&:calculate_from_report)
+          @projects = Project.order(id: :desc).limit(2)
+        end
+        format.pdf do
+          I18n.locale = 'sv'
```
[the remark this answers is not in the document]
kallenilver: Cool, thanks!

app/pdfs/climate_report_pdf.rb
kallenilver: Yes, it's intentional, emission sources only belong to one scope (while categories can have several).

kallenilver: Agree with the `.first` though!

app/pdfs/climate_report_pdf.rb
kallenilver: Since it's an Inner join it will not count the `ClimateReports` created that does not have a `ClimateReportInvoice`

app/pdfs/climate_report_pdf.rb
kallenilver: No, I don't think joining Invoice is unnecessary since it excludes `ClimateReports` without an `ClimateReportInvoice`. I can throw in calculation as well though to load all data at once. 

Accessor would be nice though, I fix.

app/pdfs/climate_report_pdf.rb
```diff
@@ -0,0 +1,199 @@
...
+    "(#{I18n.t("business.climate_reports.calculation_period_length_options.#{length}").downcase})"
+  end
+
+  # more on algorithm: https://revs.runtime-revolution.com/getting-100-with-rounded-percentages-273ffa70252b
+  # { "a": 0.5, "b": 0.51, "c": 98.9 } ->
+  # { "a": 0, "b": 1, "c": 99 } not { "a": 1, "b": 1, "c": 99 }
+  def get_even_percentages(dataset, max = 100)
+    diff = max - dataset.inject(0) { |sum, d| sum + d[1].floor }
+    dataset
+      .sort_by { |x| x[1].floor - x[1] }
```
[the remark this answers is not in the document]
kallenilver: Agree, thanks!

app/pdfs/climate_report_pdf.rb
```diff
@@ -0,0 +1,199 @@
...
+      .map
+      .with_index { |e, index| dataset[e[0]] = index < diff ? e[1].floor + 1 : e[1].floor }
```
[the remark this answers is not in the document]
kallenilver: Very 👍 , thanks! 🙏

app/pdfs/climate_report_pdf.rb
```diff
@@ -0,0 +1,199 @@
...
+    { name: 'material', scope: [3] },
+    { name: 'purchased_computers', category: 'material', scope: [3] },
+    { name: 'purchased_phones', category: 'material', scope: [3] },
+    { name: 'purchased_monitors', category: 'material', scope: [3] },
+    { name: 'other', scope: [3] },
+    { name: 'other', category: 'other', scope: [3] }
+  ].freeze
+
+  def initialize(climate_report)
+    @cr = climate_report
```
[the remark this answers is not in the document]
kallenilver: +1

app/pdfs/climate_report_pdf.rb
```diff
@@ -0,0 +1,199 @@
...
+  def categories
+    all_fields.map { |field| field unless field.key?(:category) }.compact
+  end
+
+  def emissions
+    all_fields.map { |field| field if field.key?(:category) }.compact
+  end
+
+  def all_fields
+    fields = CALCULATION_FIELDS.each { |f| f[:emissions] = @cr.calculation.send("#{f[:name]}_emissions") }
```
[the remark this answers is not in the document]
kallenilver: Ops, thank you for seeing this a bit more serious bug :) I'll fix.

app/pdfs/climate_report_pdf.rb
```diff
@@ -0,0 +1,199 @@
...
+    { name: 'flight', category: 'business_trips', scope: [3] },
+    { name: 'car', category: 'business_trips', scope: [3] },
+    { name: 'meals', scope: [3] },
+    { name: 'meals', category: 'meals', scope: [3] },
+    { name: 'material', scope: [3] },
+    { name: 'purchased_computers', category: 'material', scope: [3] },
+    { name: 'purchased_phones', category: 'material', scope: [3] },
+    { name: 'purchased_monitors', category: 'material', scope: [3] },
+    { name: 'other', scope: [3] },
+    { name: 'other', category: 'other', scope: [3] }
```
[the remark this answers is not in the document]
kallenilver: 👍

app/pdfs/climate_report_pdf.rb
kallenilver: That was my first thought, but I actually made an effort to try to do it the rails way instead :) But I agree, SQL would most likely be cleaner here.

kallenilver: @stefanlindbohm Ok, this was not as straight forward as I thought. The sum of the categories are not persisted, which makes the generation of the sql different between categories and emission sources, which in turn results in more and messy code. I vote for keeping this as is.

kallenilver: Or not as is, I'll do some clean up and simplifications.

app/pdfs/climate_report_pdf.rb
kallenilver: Agree, I fix.

app/pdfs/climate_report_pdf.rb
```diff
@@ -0,0 +1,199 @@
...
+      fields_in_category.map { |field| sources_data[field[:name]] = field_percentage(field) }
+      category_and_sources_percentages.merge!(get_even_percentages(sources_data, v))
+    end
+    fields.map { |f| f[:percentage] = category_and_sources_percentages[f[:name]] }
+    fields
+  end
+
+  def field_percentage(field)
+    BigDecimal(@cr.calculation.send("#{field[:name]}_emissions")) / @cr.calculation.total_emissions * 100
+  end
```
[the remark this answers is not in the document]
kallenilver: Agree that it belongs better in `ClimateReportCalculation`. I moved the code there and refactored it in hopes of making it easier to read.

app/pdfs/climate_report_pdf.rb
stefanlindbohm: New solution is clean indeed. Still loading every record like that is going to be a problem at some point, but since it's confined to PDF generation which doesn't happen so often it'll probably take a while until that happens. So let's just keep track of render times for PDF's and fix it when it actually is a problem we need to fix :).

app/pdfs/climate_report_pdf.rb
stefanlindbohm: Oh, right. I would prefer to write it the other way then (`ClimateReportInvoice.joins...`) to make it clear where we're starting from, but no need to change this around for now.

PR #278, opened 2020-04-20
Add simple experiments infrastructure with YML config

stefanlindbohm:
To run an experiment, just add it to `config/experiments.yml` with the desired frequency (for new sessions, resets after 30 days). Then just call `experiment_active?` with the symbolized name of the experiment to check if it is currently assigned.

We can link to experiments by using the `enable_experiments` and `disable_experiments` query params, like this: `https://www.goclimateneutral.org/?enable_experiments=new_calculator` and to disable: `https://www.goclimateneutral.org/?disable_experiments=new_calculator`. These get saved to the cookie like others so subsequent page views/visits keep that preference.

Experiments get reported to Google Analytics as a (sorted) comma separated string to the custom dimension called 'Experiments' (`dimension1`). If no experiment is active, it's reported as `none` so we can easily list the baseline separately if need be.

### DIFF
--- a/app/models/experiments_assignment.rb
+++ b/app/models/experiments_assignment.rb
@@ -1,0 +1,51 @@
+# frozen_string_literal: true
+
+class ExperimentsAssignment
+  CURRENT_EXPERIMENTS = YAML.safe_load(File.read('config/experiments.yml')).deep_symbolize_keys
+
+  attr_reader :active_experiments
+
+  def initialize(cookie_string = nil)
+    @active_experiments = []
+
+    cookie_experiments = parse_cookie_string(cookie_string)
+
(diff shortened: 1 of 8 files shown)


app/models/experiments_assignment.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def initialize(cookie_string = nil)
+    @active_experiments = []
+
+    cookie_experiments = parse_cookie_string(cookie_string)
+
+    CURRENT_EXPERIMENTS.each do |experiment, options|
+      if cookie_experiments[experiment].present?
+        @active_experiments << experiment if cookie_experiments[experiment] == '1'
+      elsif rand <= options[:frequency]
```
kallenilver, line 16: Nice, TIL about the rand method.

PR #287, opened 2020-05-08
New tailwind config

emmabckstrm:
## Todo

Verify that these pages look OK (in line with old/current design) with new config:

- [x] New calculator
- [x] Sign up
- [x] About us

### DIFF
--- a/app/assets/main/components/input.scss
+++ b/app/assets/main/components/input.scss
@@ -11,7 +11,7 @@
     border border-primary rounded bg-white
     cursor-text; /* For when we're styling a container as the input */
 
-  &:focus-within, &:focus {
+  &:focus-within, &:focus, &.StripeElement--focus {
     @apply
       border-transparent
       shadow-outline;
--- a/app/assets/main/components/input.scss
+++ b/app/assets/main/components/input.scss
@@ -6,13 +6,12 @@
   @apply
     p-2
     leading-6
-    bg-white
-    text-black
-    text-xl
-    border border-gray-400 rounded
+    text-primary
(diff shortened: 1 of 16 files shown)


app/assets/main/components/input.scss
```diff
@@ -6,13 +6,12 @@
...
-    bg-white
-    text-black
-    text-xl
-    border border-gray-400 rounded
+    text-primary
+    text-base
+    border border-primary rounded bg-white
     cursor-text; /* For when we're styling a container as the input */
 
-  &:focus-within, &:focus, &.StripeElement--focus {
```
stefanlindbohm, line 15: We want to keep the Stripe element selector

tailwind.config.js
```diff
@@ -1,55 +1,125 @@
...
-      fontSize: {
-        'big-icon': '12rem'
+    variants: {
+      borderWidth: ['responsive', 'hover']
+    },
+    screens: {
+      // Default: Mobile
+      'm-lg': '640px', // Mobile Large
+      t: '768px', // Tablet
+      'd-sm': '1024px', // Desktop Small
```
stefanlindbohm, line 10: Maybe change this to just `d` and the others to `d-lg` and `d-xl`? Usually and layout-wise, this breakpoint and up look the same. The difference will be how airy spacing will be, right? So easier to usually have `d` as a go to. I got caught by this on the other branch and wondered why the layout was wrong on the "desktop" breakpoint :).

emmabckstrm: I agree! Will fix!

PR #289, opened 2020-05-13
New design for login pages

stefanlindbohm:
Looks like this:

desktop

mobile

---

Also sets new defaults for rounded corners and drop shadows. Two of each is all you get for now, one default for smaller things (like buttons & dropdown menus) and one large for... well larger things (like callout boxes). I don't see anything else right now so let's add more when we know what they're for?

### DIFF
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -132,9 +132,9 @@ module.exports = {
         '-10': '-10'
       },
       transitionProperty: {
-        default: 'background-color, border-color, color, fill, stroke, opacity, box-shadow, transform, border-width height width max-height max-width',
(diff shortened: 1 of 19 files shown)


tailwind.config.js
```diff
@@ -98,6 +98,18 @@ module.exports = {
       'big-icon': '12rem' // temporary
     }),
     fontFamily: false,
+    borderRadius: {
+      none: '0',
+      default: '0.25rem',
+      lg: '0.5rem',
```
emmabckstrm, line 104: shouldn't this be named `md` if the step is so small we won't ever put in something in between them?

stefanlindbohm: Reason I did this is because I don't think we should have larger radii, it doesn't really fit with our visual language. But if we change our minds we can also just search and replace.

PR #302, opened 2020-06-02
Change underscore in urls to dash for seo reasons

kallenilver:
Since 'our_projects' is considered one word by search engines and 'our-projects' two words we want to replace _ with - in urls.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -57,20 +57,20 @@ Rails.application.routes.draw do
     get 'contact', to: 'welcome#contact'
     get 'faq', to: 'welcome#faq'
     get 'press', to: 'welcome#press'
-    get '100_percent_transparency', to: 'welcome#transparency', as: 'transparency'
-    get 'privacy_policy', to: 'welcome#privacy_policy'
-    get 'travel_calculator', to: 'welcome#travel_calculator'
+    get 'transparency', to: 'welcome#transparency', as: 'transparency'
+    get 'privacy-policy', to: 'welcome#privacy_policy'
+    get 'travel-calculator', to: 'welcome#travel_calculator'
 
     resources :lifestyle_footprints, path: 'calculator', only: [:new, :create], path_names: { new: '' }
 
-    resources :projects, path: 'our_projects', only: [:index]
+    resources :projects, path: 'climate-projects', only: [:index]
 
     get 'business', to: 'welcome#business'
     namespace :business do
-      resources :climate_reports, only: [:show, :new, :create], param: :key do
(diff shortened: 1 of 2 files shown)


config/routes.rb
```diff
@@ -57,20 +57,20 @@
     get 'contact', to: 'welcome#contact'
     get 'faq', to: 'welcome#faq'
     get 'press', to: 'welcome#press'
-    get '100_percent_transparency', to: 'welcome#transparency', as: 'transparency'
-    get 'privacy_policy', to: 'welcome#privacy_policy'
-    get 'travel_calculator', to: 'welcome#travel_calculator'
+    get '100-percent-transparency', to: 'welcome#transparency', as: 'transparency'
```
stefanlindbohm, line 60: Maybe simplify and do just `transparency`?

config/routes.rb
```diff
@@ -57,20 +57,20 @@
...
+    get 'privacy-policy', to: 'welcome#privacy_policy'
+    get 'travel-calculator', to: 'welcome#travel_calculator'
 
     resources :lifestyle_footprints, path: 'calculator', only: [:new, :create], path_names: { new: '' }
 
-    resources :projects, path: 'our_projects', only: [:index]
+    resources :projects, path: 'our-projects', only: [:index], as: 'projects'
```
stefanlindbohm, line 66: Maybe switch to `climate-projects`, that way it matches the title of the page ("Our Climate Projects") and would be a better match for web searches for "climate projects". What do you think?

stefanlindbohm: Also, is the `as: 'projects'` really necessary? The resource is `:projects`, which should also be the name of helpers etc?

spec/routing/redirect_routs_spec.rb
```diff
@@ -0,0 +1,100 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'Routes', type: :routing do
```
stefanlindbohm, line 5: I've never used routing specs. Nicely done! ✨

config/routes.rb
emmabckstrm: I agree!

kallenilver: Sounds good!

config/routes.rb
kallenilver: `climate-projects` sounds good, I'll change to that. Yes, you are right, helpers should automatically get namned `projects` from resource.

spec/routing/redirect_routs_spec.rb
kallenilver: Thanks!

PR #305, opened 2020-06-03
Remove old calculator (but keep LifestyleChoices model for now)

stefanlindbohm:
Keeping the model so that we can migrate all data from it to the new LifestyleCalculator/LifestyleFootprint models after new signups using this one has stopped.


stefanlindbohm:
@emmabckstrm If you have any branches on new design on sign up pages, you might have a somewhat tricky rebase. Are you ready? :)

PR #310, opened 2020-06-04
Localizations fixup

stefanlindbohm:
This turned out touching way more than I initially thought, but the cleanup ended up very nice imo. If anyone has branches that are touching locale files, let's wait for those to be merged so I can take the hit of rebasing everything (ping @emmabckstrm @kallenilver).

Notably, this PR:

- Adds a new `bin/i18n-tasks` command that helps us identify missing strings in our base locale (en), as well as unused strings in any locale. It also runs as a spec, so if we have any of those we'll notice it with a test failure immediately. No more cruft in locale files!
- Restructures locales with the idea of having separate files for each controller and per type for others. Paths match where strings are being used inside of `app/`
- Adds Crowdin configuration so we can take community translations!
- Normalizes how strings are wrapped between locales so that Crowdin won't get confused. In some cases this required code changes to adapt to some new default values.
- Also adds a "pseudo-locale", hijacking the Esperanto language code, for use with Crowdin's In-context translation feature. Try this out at https://goclimateneutral-stagin-pr-310.herokuapp.com/?translate with your Crowdin accounts (I sent invites to @kallenilver and @emmabckstrm). This will work the exact same in production because why not(?), we can then invite people to translate by simply pointing them to https://www.goclimate.com/?translate.

Crowdin is still set as a private project. My thinking is that we verify that everything works well after merging to master, then switch it to public.


emmabckstrm:
Super easy to follow when looking at it commit by commit! ✨

PR #321, opened 2020-06-08
Make controllers & actions naming more resourceful

stefanlindbohm:
I've been wanting to change this for a while. Primarily because `WelcomeController` (now removed) would just grow and grow with unrelated functionality. Nothing wrong with having lots of controllers, they don't have a cost to them. If we feel things get too cluttered, the Rails way would be to group controllers into more namespaces (like the `About` namespace I added here).

@emmabckstrm @kallenilver Are you with me? :)


kallenilver:
👍

PR #323, opened 2020-06-09
Remove LifestyleChoices

stefanlindbohm:
We're now using `LifestyleFootprint` for all calculations, so let's remove some code & DB tables!

Migrations of `LifestyleChoice`s data to `LifestyleFootprint` was done with the following migration script. `LifestyleCalculator` with ID 11 is a new calculator with all the same options as `LifestyleChoice` records for categories `food` and `car_type`, otherwise empty (it produces 0 for all categories if you use it to calculate anything now). It has version number 1, lower than any of the published ones, so it will never be used for any new calculations.

```ruby
legacy_calculator = LifestyleCalculator.find(11)

User.joins(:lifestyle_choices).includes(:lifestyle_choices).find_each do |u|
  choices = u.lifestyle_choices.map do |choice|
    [
      choice.category == 'custom' ? :flying : choice.category.to_sym,
      { choice: choice.name, co2e: (choice.co2 * 1000).to_i}
    ]
  end.to_h

  f = LifestyleFootprint.new(
    user: u,
    lifestyle_calculator: legacy_calculator,
    created_at: u.created_at,
    housing: choices[:base][:co2e],
    food_answer: choices[:food][:choice],
    food: choices[:food][:co2e],
    car_type_answer: choices[:car][:choice],
    car: choices[:car][:co2e],
    flights: choices[:flying][:co2e],
    consumption: 0,
    public: 0
  )

  f.total = f.housing + f.food + f.car + f.flights

  f.save!(validate: false)
end
```

Backups of both dropped tables as well as the above migration script are in our Drive folder under `Product/Backups`.


app/models/climate_report_invoice.rb
```diff
@@ -7,8 +7,10 @@ class ClimateReportInvoice < ApplicationRecord
   validates_presence_of :invoice_address, :co2e, :amount, :currency
   validates :certificate_reciever_email, email: true
 
+  BUFFER_FACTOR = 2
```
kallenilver, line 10: This buffer should be same for business and individuals, so if it's possible to define it in one place for both uses that would be neat I guess.

app/models/currency.rb
```diff
@@ -27,6 +27,3 @@ def suffix
 Currency::SEK = Currency.new(:sek)
 Currency::EUR = Currency.new(:eur)
 Currency::USD = Currency.new(:usd)
-
-Currency::SEK_PER_USD = 8.5
-Currency::SEK_PER_EUR = 10
```
kallenilver, line 32: I agree that this does not belong in currency, GHG is a better choice. Nice!

kallenilver:
@stefanlindbohm Regarding the migration, `f.total` does not include the number of persons you offset for?

app/models/climate_report_invoice.rb
stefanlindbohm: First off it isn't used by both as `SubscriptionManager` (which does the pricing for subscriptions) defines a factor of 2 on its own. So I moved the constant to the one place it is actually used. Same for `GiftCard` (even though that code definitely could get some cleanup).

I thought about it too and my conclusion is that while it's true that we currently have the same buffer factor, it's not necessarily so that we'll always do. In fact, we used to have no buffer for `ClimateReportInvoice`s. So I landed on "don't make abstractions until there is a clear pattern". A constant isn't much to repeat anyway and it makes code much easier to read.

stefanlindbohm:
> @stefanlindbohm Regarding the migration, `f.total` does not include the number of persons you offset for?

Correct. The new calculator & `LifestyleFootprint` have no concept of number of people, by design. The calculation itself is for one person. Number of people now only multiplies the calculated footprint to create a subscription of the equivalent amount. Basically just as a quick and understandable way to choose another subscription amount, similar to what you can do by going to billing settings when logged in. My reasoning is that number of people carries no real meaning–it's the subscription amount that does, which can be changed independently of the calculated footprint anyway.

PR #329, opened 2020-06-12
New new design signup

emmabckstrm:


app/assets/main/components/button.scss
```diff
@@ -5,7 +5,7 @@
     border rounded border-current
     bg-transparent
     whitespace-no-wrap
-    text-primary
+    text-primary text-center
```
kallenilver, line 8: Newbie-question, I take it that it's our coding standard to have class names that concern the same thing (like text in this case) on the same row?

app/assets/main/components/callout.scss
```diff
@@ -1,5 +1,6 @@
 .callout {
-  @apply bg-white rounded-lg shadow-lg p-4;
+  @apply bg-white rounded-lg shadow-lg p-4
+  border border-gray-tint-2;
```
kallenilver, line 3: And that indentation after `@apply `is only needed if `@apply` is on an own row?

app/assets/main/components/button.scss
stefanlindbohm: Yep. We haven't codified it but this is what we started doing in practice :)

app/assets/main/components/callout.scss
stefanlindbohm: Nah, this one I think is wrong. Always indent!

app/views/devise/registrations/new_new.html.erb
```diff
@@ -1,162 +1,200 @@
...
-        </button>
-        <p class="text-orange-shade-1 ml-4" data-target="registration-form.errorMessage"></p>
+
       </div>
     </div>
+  </div>
+</section>
+
+<div class="bg-blue-tint-1 relative">
+  <%= image_tag webpack_asset_path('images/illustrations/forest_and_city.png'), alt: 'Forest and city illustration', class: 't:max-w-3/4 mx-auto' %>
```
stefanlindbohm, line 171: I think this should be limited to its size as in the design file, i.e. `class="w-3/4 mx-auto" style="max-width: 812px"`. It gets fuzzy and unnecessarily big on bigger screens now.

app/views/devise/registrations/new_new.html.erb
```diff
@@ -1,162 +1,200 @@
...
+  <%= image_tag webpack_asset_path('images/illustrations/horizon_curve.svg'), alt: 'Curve illustration', class: 'w-full absolute bottom-0 -mb-1' %>
```
stefanlindbohm, line 172: I like the touch of adding `alt`s to images, but these are purely decoration so maybe these instead be hidden from screen readers? The current texts would be slightly confusing to hear, right?

PR #339, opened 2020-06-22
Add new design for contact us page

stefanlindbohm:
This one is simple.

Screen Shot 2020-06-22 at 12 23 11

### DIFF
--- a/app/views/about/contact/show_new.html.erb
+++ b/app/views/about/contact/show_new.html.erb
@@ -1,0 +1,13 @@
+<%= render "shared/header" %>
+
+<section class="section-padding">
+  <h1 class="heading-xl mb-6"><%= title(t('contact_us')) %></h1>
+  <div class="max-w-2xl space-y-6">
+    <p><%= description(t('general_questions_html')).gsub('<a', '<a class="link" rel="noopener noreferrer"').html_safe %></p>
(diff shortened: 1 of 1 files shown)


app/views/about/contact/show_new.html.erb
```diff
@@ -0,0 +1,13 @@
+<%= render "shared/header" %>
+
+<section class="section-padding">
+  <h1 class="heading-xl mb-6"><%= title(t('contact_us')) %></h1>
+  <div class="max-w-2xl space-y-6">
+    <p><%= description(t('general_questions_html')).gsub('<a', '<a class="link"').html_safe %></p>
+    <p><%= t('general_questions_press_html').gsub('<a', '<a class="link"').html_safe %></p>
+    <p><%= t('general_questions_facebook_html').gsub('<a', '<a class="link"').html_safe %></p>
+    <p><%= t('general_questions_twitter_html').gsub('<a', '<a class="link"').html_safe %></p>
```
emmabckstrm, line 9: Nice substitution. When you're at it, I think it's good to add `rel="noopener noreferrer"` to the `a` tags that are opened in new tabs and goes to external sites, as well
https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/

stefanlindbohm: Wow, I did not know about `window.opener`. Adding!

PR #342, opened 2020-06-22
Add logos and links to swedish media we have been featured in

emmabckstrm:
## New design 
Screenshot 2020-06-22 at 14 03 45
Screenshot 2020-06-22 at 14 03 57
Screenshot 2020-06-22 at 14 04 09

---

## Old design

Screenshot 2020-06-22 at 14 01 14
Screenshot 2020-06-22 at 14 01 29

### important

The code for the old design is neither pretty, nor perfect, but going back to writing Bootstrap CSS from being used to Tailwind is .... terrible


app/views/about/show.html.erb
```diff
@@ -94,8 +93,58 @@
...
+              url: "https://tidningensyre.se/stockholm/leva-ett-klimatneutralt-liv/"
+            }
+          ].each_with_index do |logo, index| %>
+            <% if index == 2 || index == 4 %>
+              </div>
+            <% end %>
+            <% if index == 0 || index == 2 || index == 4 %>
+              <div class="row">
+            <% end %>
+            <div class="col-xs-6 col-sm-6 text-center" style="height: 120px; margin: 10px auto;">
```
emmabckstrm, line 134: manual styling here because Bootstrap is a pain, especially when I'm being used with Tailwind

app/views/about/show_new.html.erb
```diff
@@ -104,6 +104,44 @@
...
+            image_url: "veckansaffarer-logga_low_300.png",
+            url: "https://www.va.se/nyheter/2019/07/16/vi-maste-konsumera-mycket-mindre/"
+          },
+          {
+            image_url: "syre.png",
+            url: "https://tidningensyre.se/stockholm/leva-ett-klimatneutralt-liv/"
+          }
+        ].each do |logo| %>
+          <a href="<%= logo[:url] %>" target="_blank" rel="noopener" class="w-1/2 md:w-1/3 flex items-center justify-center p-10" style="max-height: 120px">
+            <%= image_tag webpack_asset_path('images/press_logos/' + logo[:image_url]), class: 'max-h-full' %>
```
stefanlindbohm, line 138: For simplicity, I'd put the full path in the above hash directly, instead of having to look in two places to see the full thing. Also avoids interpolation & makes it easier to search for (if you're searching for the full path).

app/views/about/show.html.erb
stefanlindbohm: I don't care the least about the old templates 😎

stefanlindbohm:
Oh wait, why is this in between the two text blocks? Shouldn't it be after?

emmabckstrm:
> Oh wait, why is this in between the two text blocks? Shouldn't it be after?

@stefanlindbohm what do you mean? I can't see what the comment is referencing

stefanlindbohm:
> > Oh wait, why is this in between the two text blocks? Shouldn't it be after?
> 
> @stefanlindbohm what do you mean? I can't see what the comment is referencing

I mean the section of logos being after "Why does GoClimate exist?" and before "How do we stop climate change?". To me, those two are more connected and the press logos should be either before or after both of them.

emmabckstrm:
> > > Oh wait, why is this in between the two text blocks? Shouldn't it be after?
> > 
> > 
> > @stefanlindbohm what do you mean? I can't see what the comment is referencing
> 
> I mean the section of logos being after "Why does GoClimate exist?" and before "How do we stop climate change?". To me, those two are more connected and the press logos should be either before or after both of them.

@stefanlindbohm Ah, I see. Hm, probably legacy from an older discussion, which I don't really remember the reason, so I'll move it down!

PR #343, opened 2020-06-22
Updates from content map

stefanlindbohm:
- Reorders main nav items to this order: Projects, About, Gift Cards, For Business. Projects is slightly more visited than About, hence that ordering.
- Tweaks margins in main nav & prefooter
- Makes button in prefooter a main CTA (yellow background)
- Tweaks alignment on about page
- Links between about, project & transparency pages.


app/views/about/show_new.html.erb
```diff
@@ -18,10 +18,10 @@
 </section>
  
 <section class="section-padding">
-  <h2 class="heading-lg text-center mb-12">
+  <h2 class="heading-lg mb-12">
     <%=t 'about_us_who_are_we' %>
   </h2>
-  <div class="flex flex-wrap justify-center section-gutter">
```
emmabckstrm, line 24: Hmmmmmm, okay then. I personally like the centered version but this is okay.
Though, it looks a bit silly on really big screens as the `Who are we?` section becomes wider than `About us` and `Why does GoClimate exist?`

PR #352, opened 2020-07-02
Replace Font Awesome icons with Material icons

emmabckstrm:
## Let's keep this PR open until it's been thoroughly discussed

Stylewise, this is a change we want to do. However, there are some differences in implementing the icons that could be taken into account, and also the reason for leaving this as a PR without merging.

### Things to consider

1. Material icons align differently with text than FA icons, which requires additional markup to solve. As can be seen in the changes, I've done it with adding `flex items-center` to the parent element.
2. Sizing. Material Icon's base size is 24px, and is set through the icon font class, which means it's unaffected by utility classes that affect font size, without CSS fixes. [Material icons](https://google.github.io/material-design-icons/#styling-icons-in-material-design) suggest size steps of either 18, 24, 36 or 48px. 
3. Not all Material icons fill up their space in the same way, making it difficult and honestly pretty frustrating to align them. For instance, I was about the change the quote as seen below. There were basically no issues when using FA, but here with Material icons I have to dab around with minus margin to make the icon line up with the text to the left.
Screenshot 2020-07-10 at 10 23 53
4. As Material Icons doesn't include social media icons, social icons have not been replaced, and we either need to keep Font Awesome or implement some other solution to handle social media icons.

In summary, I'm actually leaning towards NOT going over to Material icons, mainly due to the overhand in actually using them. To me, we use icons so rarely that they are barely visible, and then there is not a big reason for me to a) add more to our bundle size and b) slow down the implementation of icons. 

### Notes

- Where icons have been used in templates with old designs, the icons have not been replaced with Material icons

### DIFF
--- a/app/assets/main/components/lists.scss
+++ b/app/assets/main/components/lists.scss
@@ -13,8 +13,8 @@
 
 .list-check {
   li::before {
-    font-family: FontAwesome;
-    content: "\f00c";
+    font-family: "Material Icons Round";
+    content: "check";
     font-size: 1.5rem;
 
     @apply
--- a/app/assets/main/components/select.scss
+++ b/app/assets/main/components/select.scss
@@ -18,8 +18,8 @@
   @apply relative inline-block;
 
   &::after {
-    font-family: FontAwesome;
-    content: "\f107";
+    font-family: "Material Icons Round";
+    content: "expand_more";
 
     @apply 
       absolute right-0 top-1/2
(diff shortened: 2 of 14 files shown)


app/assets/main/components/lists.scss
```diff
@@ -13,8 +13,8 @@
 
 .list-check {
   li::before {
-    font-family: FontAwesome;
-    content: "\f00c";
+    font-family: "Material Icons Round";
+    content: "check";
     font-size: 1.5rem;
```
emmabckstrm, line 18: Works perfectly as this evaluates to 24px, which is the standard size for material icons

PR #353, opened 2020-07-03
Re-add Meet the community to startpage

emmabckstrm:
Screenshot 2020-07-06 at 14 08 08
Screenshot 2020-07-06 at 14 10 07

Notes

- Fixed height on the container box as the texts for different people are different in length and makes the page jump around when switched.
- Had to add new heights to tailwind as I needed them to be used in different breakpoints

Todo

- [x] Get green light on swedish translation of Lydia's text


stefanlindbohm:
👏

PR #356, opened 2020-07-09
New design settings

emmabckstrm:
## Notes

The general design suggestion is pretty much there, but the functionality for payment settings is not finished. And I probably don't need to mention but the design shown in the screenshot for the Payment settings is not completely finished on a detailed level.

Screenshot 2020-07-09 at 10 04 49

Screenshot 2020-08-20 at 15 45 25

Screenshot 2020-08-20 at 15 50 05


app/views/devise/registrations/edit_old.html.erb
```diff
@@ -0,0 +1,71 @@
+<% content_for :stylesheets do %>
```
stefanlindbohm, line 1: This is not needed since we migrated fully to the new design?

app/views/users/subscriptions/show_old.html.erb
```diff
@@ -0,0 +1,79 @@
+<% content_for :stylesheets do %>
```
stefanlindbohm, line 1: Also not needed.

stefanlindbohm (approved):

PR #361, opened 2020-08-12
Co2e per month in admin dashboard

kallenilver:
This would be easier to do with a refactoring to log sold co2e (and not just amount paid) for `CardCharges` in place. But this works until that is in place.

### DIFF
--- a/app/models/card_charge.rb
+++ b/app/models/card_charge.rb
@@ -42,6 +42,32 @@ class CardCharge < ApplicationRecord
       total_payments_eur_part * GreenhouseGases::PRICE_FACTOR_EUR).round
   end
 
+  def self.payments_per_month
+    where(paid: true)
+      .group("CONCAT((EXTRACT(YEAR FROM created_at)), '-', EXTRACT(MONTH FROM created_at))")
+      .group('currency')
+      .order('2 DESC')
+      .sum('amount / 100')
+  end
+
+  def self.co2e_per_month
+    co2 = {}
+    payments_per_month.each do |(month, currency), amount|
+      sek_amount =
+        case currency
+        when 'usd'
+          amount * GreenhouseGases::PRICE_FACTOR_USD
+        when 'eur'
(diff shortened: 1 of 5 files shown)


app/models/card_charge.rb
```diff
@@ -42,6 +42,23 @@ def self.total_in_sek
...
+           .sum('amount / 100')
+    co2 = {}
+    paid.each do |month, amount|
+      sek = amount * GreenhouseGases::PRICE_FACTOR_USD if month[1] == 'usd'
+      sek = amount * GreenhouseGases::PRICE_FACTOR_EUR if month[1] == 'eur'
+      sek = amount if month[1] == 'sek'
+      co2[month[0]] = 0 if co2[month[0]].nil?
+      co2[month[0]] += sek / GreenhouseGases::CONSUMER_PRICE_PER_TONNE_SEK.amount
+    end
+    co2
```
stefanlindbohm, line 59: I had some trouble understanding this because of variable names and magic indices. I'd suggest something like this for clarity. (Untested, so please test against your expected behavior.)

```ruby
    paid = where(paid: true)
           .group("EXTRACT(YEAR FROM created_at) || '-' || EXTRACT(MONTH FROM created_at)")
           .group('currency')
           .order('2 DESC')
           .sum('amount / 100')

    co2 = {}
    paid.each do |(month, currency), amount|
      sek_amount =
        case currency
        when 'usd'
          amount * GreenhouseGases::PRICE_FACTOR_USD
        when 'eur'
          amount * GreenhouseGases::PRICE_FACTOR_EUR
        when 'sek'
          amount
      end
      
      co2[month] = 0 if co2[month].nil?
      co2[month] += sek_amount / GreenhouseGases::CONSUMER_PRICE_PER_TONNE_SEK.amount
    end

    co2
```

kallenilver: Looks great, thanks!

PR #362, opened 2020-08-17
Update business report

kallenilver:
New emission factors + adding coffee and fruit emissions to other category.


stefanlindbohm (approved):
Looking good!

PR #364, opened 2020-08-21
New design on error pages

emmabckstrm:
Super quick and easy translation of new design onto error pages.

Screenshot 2020-08-21 at 09 22 34

Screenshot 2020-08-21 at 09 22 08
Screenshot 2020-08-21 at 09 22 26

### DIFF
--- a/app/views/errors/internal_server_error.html.erb
+++ b/app/views/errors/internal_server_error.html.erb
@@ -1,12 +1,12 @@
 <%= render "shared/header" %>
-  <section class="section-padding text-center">
-    <div class="callout">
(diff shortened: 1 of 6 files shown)


stefanlindbohm:
I was itching a little over the long centered lines of text and the use of the callout box that to me doesn't really match how we use it for the rest of the site (for actionable stuff/more important info than other on the page). What do you think about this?

While I was at it, I made it so that the area below the footer is always the same background color as the footer.

Screen Shot 2020-08-24 at 17 48 18

Screen Shot 2020-08-24 at 17 45 10

emmabckstrm:
Looks fine! Agree on the use of callout box! @stefanlindbohm

PR #370, opened 2020-08-26
Remove card when terminating subscription

emmabckstrm:


app/models/subscription_manager.rb
```diff
@@ -37,6 +37,17 @@ def cancel
     subscription.delete
   end
 
+  def remove_payment_methods
+    if customer.invoice_settings&.default_payment_method
+      payment_methods = Stripe::PaymentMethod.list({ customer: customer.id, type: 'card' })
+      payment_methods.data&.each do |payment_method|
```
stefanlindbohm, line 43: If you skip `.data`, you should get `Stripe::PaymentMethod` instances directly, I think?

app/models/subscription_manager.rb
```diff
@@ -37,6 +37,17 @@ def cancel
...
+    if customer.invoice_settings&.default_payment_method
+      payment_methods = Stripe::PaymentMethod.list(customer: customer.id, type: 'card')
+      payment_methods.each do |payment_method|
+        Stripe::PaymentMethod.detach(payment_method.id)
+      end
+    elsif customer.default_source
+      customer.sources.detach_source(customer.default_source)
+    end
+  end
+
```
stefanlindbohm, line 50: I feel like this should have unit tests. Added benefit is that tests then will also explain why the branching is happening (why are we looking for both `invoice_settings.default_payment_method` and `customer.default_source`.

stefanlindbohm (commented):
It's all looking dandy dandy, but I do miss one unit test here.

app/controllers/users/subscriptions_controller.rb
```diff
@@ -65,11 +65,9 @@ def customer_currency
...
 
     def customer_payment_method(customer)
-      if customer.invoice_settings&.default_payment_method
-        Stripe::PaymentMethod.retrieve(customer.invoice_settings.default_payment_method)
-      elsif customer.default_source
-        customer.sources.retrieve(customer.default_source)
-      end
+      return nil unless customer.invoice_settings&.default_payment_method
+
+      Stripe::PaymentMethod.retrieve(customer.invoice_settings.default_payment_method)
```
emmabckstrm, line 70: `PaymentMethod` API throws error if you're trying to access an `id` that is `nil`

stefanlindbohm (approved):
Very nice!

PR #373, opened 2020-09-01
Use new modal component for cancelling subscription

emmabckstrm:
Screenshot 2020-09-01 at 14 28 59

### DIFF
--- a/app/views/users/subscriptions/show.html.erb
+++ b/app/views/users/subscriptions/show.html.erb
@@ -95,9 +95,24 @@
     <% if @current_plan_price %>
       <div class="callout max-w-xl mx-auto">
         <h2 class="heading mb-3"><%= t 'views.subscriptions.cancel.heading' %></h2>
-        <%= form_tag({ action: :destroy }, method: :delete) do %>
-          <%= button_tag t('views.subscriptions.cancel.action'), class: "button w-full", onclick: "return confirm('#{t('views.subscriptions.cancel.confirm')}')" %>
-        <% end %>
+        <div data-controller="modal">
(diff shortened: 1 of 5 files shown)


app/views/users/subscriptions/show.html.erb
```diff
@@ -93,11 +93,25 @@
...
+        <div class="modal-container" data-target="modal.container">
+          <div class="modal-overlay"></div>
+          <div class="modal-content" data-target="modal.content">
+            <h3 class="heading-lg mb-3"><%= t('views.subscriptions.cancel.action') %></h3>
+            <%= t('views.subscriptions.cancel.confirm') %>
+            <div class="modal-actions">
+              <%= form_tag({ action: :destroy }, method: :delete, class: 'button') do %>
+                <%= button_tag t('views.subscriptions.cancel.action'), id: 'confirm_cancellation' %>
+              <% end %>
+              <button class="button button-cta"><%= t('views.subscriptions.cancel.abort') %></button>
```
stefanlindbohm, line 111: I'm influenced by Apple's design guidelines here, but I feel strongly that the "continue" or "primary" action should be the rightmost button in modals. I.e. the "Cancel subscription" button should be to the right. It's the "most continue" option and continue is intuitively to the right.

Also, since this is a confirmation on the intent to cancel a subscription, I think that button should be the one to have the CTA color. Or in this case, since it's destructive, maybe we could figure out an alternate destructive color to use? Maybe it's time to find a good red to use specifically for destructive/error in our UI's.

stefanlindbohm (commented):
Implementation is super clean!

PR #374, opened 2020-09-03
Offboarding

emmabckstrm:
Of course should be merged into master but doing it like this makes it really easy to see which commits are actually pushed in this branch.

Screenshot 2020-09-03 at 11 36 13
Screenshot 2020-09-03 at 11 36 31

### DIFF
--- a/app/controllers/users/subscriptions_controller.rb
+++ b/app/controllers/users/subscriptions_controller.rb
@@ -31,8 +31,17 @@ module Users
 
     def destroy
       @manager = SubscriptionManager.new(@stripe_customer)
+
+      cancellation_feedback = SubscriptionCancellationFeedback.new(
+        subscribed_at: Time.at(@manager.subscription.start_date),
+        reason: params[:cancellation_reason] == 'other' ?
+            params[:cancellation_reason_other_text] : params[:cancellation_reason]
+      )
+      cancellation_feedback.save
+
       @manager.cancel
       @manager.remove_payment_methods
+
       redirect_to action: :show
       flash[:notice] = I18n.t('views.subscriptions.cancel.cancellation_successful')
     end
--- a/app/models/subscription_cancellation_feedback.rb
+++ b/app/models/subscription_cancellation_feedback.rb
@@ -1,0 +1,4 @@
+# frozen_string_literal: true
+
+class SubscriptionCancellationFeedback < ApplicationRecord
+end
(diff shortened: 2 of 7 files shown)


db/migrate/20200902_create_subscription_cancellations.rb
```diff
@@ -0,0 +1,10 @@
+class CreateSubscriptionCancellations < ActiveRecord::Migration[6.0]
```
emmabckstrm, line 1: I wrote this file manually. I'm not sure about the name of the file, the others seem to have more numbers after the date, but couldn't figure out what they were so leaving it like this and asking what to do :D

app/models/subscription_cancellation.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+class SubscriptionCancellation < ApplicationRecord
```
stefanlindbohm, line 3: I feel this is not really really representing the cancellation itself (that is in Stripe). What about `SubscriptionCancellationFeedback`?

db/migrate/20200902_create_subscription_cancellations.rb
stefanlindbohm: I think it's just the full timestamp. Best is to generate these files with `bin/rails generate migration CreateSubscriptionCancellation`

db/migrate/20200902_create_subscription_cancellations.rb
```diff
@@ -0,0 +1,10 @@
...
+  def change
+    create_table :subscription_cancellations do |t|
+      t.timestamps
+
+      t.timestamp :signed_up
```
stefanlindbohm, line 6: To be in line with other timestamp fields, I propose this gets named `signed_up_at`.

app/controllers/users/subscriptions_controller.rb
```diff
@@ -30,9 +30,16 @@ def update
     end
 
     def destroy
+      cancellation = SubscriptionCancellation.new(
+        signed_up: current_user.created_at,
```
stefanlindbohm, line 34: In cases where someone resubscribed, this won't be correct. But you can get to the actual signup date for the current subscription if you move `SubscriptionManager` initialization up and getting it through `@manager.subscription.start_date`.

stefanlindbohm (commented):
Looking good!

db/migrate/20200902_create_subscription_cancellations.rb
emmabckstrm: I gave you one better even, with `subscribed_at`

PR #378, opened 2020-09-07
New design on For Business page

emmabckstrm:
It's far from perfect but it's better than the old design :) 

The callout box around the logos is because some of the logos are screenshots against a white background, so I took this shortcut.

https://goclimateneutral-stagin-pr-378.herokuapp.com/business


stefanlindbohm (approved):
Agree it's better than what we have :)

PR #381, opened 2020-09-08
New design travel calculator

emmabckstrm:
https://goclimateneutral-stagin-pr-381.herokuapp.com/travel-calculator


stefanlindbohm (commented):
I'm suggesting we left-justify this, because those bars and labels go all kinds of directions when we try to center them. It's also more in line with other pages using our new design.

PR #382, opened 2020-09-08
New design receipts

emmabckstrm:
I also added a link to the receipts on the payment settings page.

https://goclimateneutral-stagin-pr-382.herokuapp.com/

### DIFF
--- a/app/views/users/receipts/index.html.erb
+++ b/app/views/users/receipts/index.html.erb
@@ -1,44 +1,27 @@
-<% content_for :stylesheets do %>
-  <%= webpack_entrypoint_stylesheet_tags 'main_old' %>
-<% end %>
-<% content_for :javascripts do %>
-  <%= webpack_entrypoint_javascript_tags 'main_old', onerror: 'reportLoadError(event)' %>
-<% end %>
+<%= render "shared/header" %>
 
-<div class="site-wrapper-first">
-  <%= render "shared/header_old" %>
-  <div class="inner">
+<div class="bg-blue-pastel">
+  <section class="section-padding">
+    <h1 class="heading-xl mb-6"><%= t 'views.receipts.heading' %></h1>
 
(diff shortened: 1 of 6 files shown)


app/views/users/receipts/index.html.erb
```diff
@@ -1,44 +1,27 @@
...
-
-  </div>
-  <div class="row">
-
-    <h1>Your receipts</h1>
-    <div class="col-xs-12">
-      <p class="lead">Here you can see your receipts</p>
     </div>
+  </section>
+  <%= render "shared/bottom_landscape" %>
```
stefanlindbohm, line 23: Do we have a principle on what pages we use this, actually? I thought it was more of a home page + sign up flows kind of deal?

config/locales/views/receipts.sv.yml
```diff
@@ -0,0 +1,8 @@
+sv:
+  views:
+    receipts:
+      heading: Dina kvitton
+      date: Datum
+      amount: Belopp
+      link: Länk
+      view: Visa kvitto
```
stefanlindbohm, line 8: Nice :+1:

stefanlindbohm (approved):
Actual receipt PDF is kind of schizophrenic still, but good enough for now!

app/views/users/receipts/index.html.erb
emmabckstrm: Not that we have written down, no. But I've kept on using it on logged in pages as the bottom landscape is so pretty and nice and gives character and life (especially since we have basically no other graphical elements), BUT it makes no sense to have a sign up CTA on logged in pages so I've been using this approach.

emmabckstrm: @stefanlindbohm

stefanlindbohm: Maybe instead of `skip_prefooter`, what we want is `skip_call_to_action` then? 🤔 And then consistently using it on all pages.

PR #385, opened 2020-09-09
New design change password

emmabckstrm:
Same design as `Log in` and `Forgot password` basically :blush:  

Screenshot 2020-09-09 at 09 07 59


kallenilver (approved):
Very clean! 💯

PR #388, opened 2020-09-15
Small bugs on profile page

emmabckstrm:
Learning: don't use background colours on elements with `section-padding` class


stefanlindbohm (approved):
:shipit:

PR #397, opened 2020-09-30
Categories on signup

emmabckstrm:
Screenshot 2020-10-02 at 15 27 12

Screenshot 2020-10-02 at 15 27 19
Screenshot 2020-10-02 at 14 57 45
Screenshot 2020-10-02 at 14 57 52
Screenshot 2020-10-02 at 14 58 00

### DIFF
--- a/app/views/devise/registrations/new.html.erb
+++ b/app/views/devise/registrations/new.html.erb
@@ -8,133 +8,193 @@
 
     <div class="section-gutter flex flex-col t:flex-row t:justify-between">
       <div class="space-y-6 d:space-y-12 t:w-1/2">
-        <div>
-          <h2 class="text-sm"><%= t 'dashboard.footprint.heading' %></h2>
-          <div class="mt-1">
-            <span class="text-xl font-bold text-green-accent"><%= @footprint_tonnes.tonnes.round(1) %></span>
-            <span class="text-lg font-semibold"><%= t 'tonnes_CO2' %></span>
+        <% if @active_experiments&.include?(:new_signup) %>
+          <div>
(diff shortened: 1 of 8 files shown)


app/views/devise/registrations/new.html.erb
```diff
@@ -2,67 +2,139 @@
...
-                </div>
-              </div>
-              <div>
-                <p class="relative z-10 "><%= t 'views.registrations.goal_2030' %></p>
-                <div class="relative h-6 bg-primary rounded-r" style="width: <%= 2.5 / max_value * 100 %>%">
-                  <span class="absolute left-100 ml-2 font-bold">2.5</span>
+        <div data-controller="tabs">
+          <ul data-target="tabs.tablist">
+            <li>
+              <a href="#" data-target="tabs.tab"><%= t 'dashboard.footprint.heading_compared' %></a>
```
stefanlindbohm, line 22: I feel linking to `#` is a bit unnecessary. If we're not using the anchor functionality it might as well be a `<span>`, or alternatively we can rely on the page anchor for switching tabs (also making it possible to link to the page with a specific tab open). The latter is more work so maybe the former for now?

stefanlindbohm (approved):
I like it!

Maybe have a round with non-programmer team mates on the copy & etc before merging straight to production? Also, are we interested in an A/B test for this or do we just go? Hehe

app/views/devise/registrations/new.html.erb
emmabckstrm: Made it `<button>` instead, as it's a clickable element.

PR #398, opened 2020-10-01
Add feature specs for editing subscription with and w/o 3D Secure card

stefanlindbohm:
Ping @kallenilver @emmabckstrm, these changes will be good to know about.


emmabckstrm (approved):
👏

PR #400, opened 2020-10-06
Template for sign up without subscription

emmabckstrm:
Edit:

I have separated "change membership" into its own controller. It's still a Javascript controller because the user flow we want is actually pretty complicated.


app/assets/main/controllers/registration_form_controller.js
```diff
@@ -91,6 +148,26 @@ export default class RegistrationFormController extends Controller {
...
+  }
+};
+
+RegistrationFormController.styleInactiveChoice = function styleInactiveChoice(target) {
+  target.parentNode.classList.add('bg-gray-pastel');
+  target.parentNode.classList.remove('bg-green-tint-1');
+  if (target.checked) {
+    target.checked = false;
+  }
+};
```
stefanlindbohm, line 171: Here's an example of the above comment. I've done this in the past with putting those classes in `data` attributes, such as `data-active-classes="..."` and `data-inactive-classes=""` to keep all styling in the markup. Saves lines of JS too.

app/assets/main/controllers/registration_form_controller.js
```diff
@@ -5,6 +5,63 @@ export default class RegistrationFormController extends Controller {
...
+    });
+  }
+
+  showNoSubscriptionInfo() {
+    this.noSubscriptionInfoTargets.forEach((target) => {
+      target.classList.remove('hidden');
+    });
+    this.subscriptionInfoTargets.forEach((target) => {
+      target.classList.add('hidden');
+    });
```
stefanlindbohm, line 64: I feel this should be its own Stimulus controller, and also that some of the DOM manipulation could instead be just markup. I know this is the topic we talked about yesterday and that maybe we have a little bit of a different model of this, so maybe we should talk about it and look at it together?

app/assets/main/controllers/registration_price_controller.js
```diff
@@ -1,10 +1,23 @@
...
+    const formData = Array.from(new FormData(this.peopleFormTarget));
+    const updatedFormData = formData.map(
+      (param) => (param[0] === 'people' ? ['people', people] : param)
+    );
+    const url = `${this.peopleFormTarget.dataset.url}?${serializeQueryParams(updatedFormData)}`;
+    fetch(url, {
+      method: 'get',
+      credentials: 'include',
+      headers: { Accept: 'application/json' }
+    })
```
stefanlindbohm, line 20: Is this so that the dropdown input field is not used when the "multiple people" option is not selected? If so, I think a much cleaner solution would be to disable that select field and have a hidden field before it in the HTML with the default value of 1. That way we don't need to have our own magic form posting logic here.

app/assets/util/submit_form.js
```diff
@@ -1,5 +1,5 @@
-function serializeQueryParams(formData) {
-  return Array.from(formData).map(
+export function serializeQueryParams(formData) {
+  return formData.map(
```
stefanlindbohm, line 2: If this is to be an exported function, it should have its own file here in `util/` (or be part of a file with a name that's not specific to submitting forms). But I'm not sure it should be exported, see above.

app/controllers/users/registrations_controller.rb
```diff
@@ -15,6 +15,8 @@ def new
       @projects = Project.order(id: :desc).limit(3)
       @country_average = LifestyleFootprintAverage.find_by_country(@footprint.country)
 
+      @sign_up_without_subscription = params[:sign_up_without_subscription].present?
```
stefanlindbohm, line 18: This doesn't have to be here in the controller as it's just a param that gets read and nothing happens to. I'd just put `params[:sign_up_without_subscription]` straight in the template.

app/views/devise/registrations/new.html.erb
```diff
@@ -197,101 +197,221 @@
...
-                  <%= form_with method: :get, html: { 'data-target': 'registration-price.peopleForm' } do |f| %>
-                    <%= f.hidden_field(:lifestyle_footprint, :value => @footprint.key) %>
-                    <%= f.select :people,
-                      options_for_select(1..20, params[:people].presence || 1),
-                      {},
-                      { class: 'select', 'data-action': 'registration-price#update' }
-                    %>
-                  <% end %>
+      <div class="t:w-1/2">
+        <% if @sign_up_without_subscription %>
```
stefanlindbohm, line 201: How about a partial for each of the versions of this form? Just to keep us a bit saner while we do this.

app/assets/main/controllers/registration_price_controller.js
stefanlindbohm: Or maybe even just post the radio buttons as well and let the backend handle it? Depending on what makes sense in the markup.

app/assets/main/controllers/registration_form_controller.js
emmabckstrm: Fixed!

stefanlindbohm (approved):
Sparkles ✨

PR #401, opened 2020-10-08
Set canonical to index action for the different new-versions

kallenilver:


stefanlindbohm (approved):
I'm in agreement with CodeClimate, otherwise 👍

PR #402, opened 2020-10-09
Sign up without subscription

emmabckstrm:


config/locales/devise.views.en.yml
```diff
@@ -82,6 +82,7 @@ en:
       new:
         sign_up: Sign up
       signed_up: Welcome to a climate neutral life!
+      signed_up_without_subscription: Welcome to GoClimate!
```
emmabckstrm, line 85: As they don't have a subscription, their lives are not climate neutral 😅 sidenote: perhaps we should say something else than "climate neutral life" here anyways

PR #403, opened 2020-10-13
Sign up without subscription 3

emmabckstrm:
Commits , 660084b, 6b2dbd1 and 3674412 are those changes that are actually in this PR. 

I will clean up commits when it comes time to merge, but for now I have wrestled with this long enough before making it reviewable. Will merge `sign-up-for-free-tempalte` into master before cleaning this one up to keep my sanity.

edit: see summary here https://github.com/goclimateneutral/goclimateneutral/pull/403#issuecomment-709287333
~~Also, there's a TODO regarding tracking that will need to be fixed before final reveal.~~

~~Note that this feature will be behind feature flag AND A/B test.~~

### DIFF
--- a/app/assets/main/controllers/membership_choice_controller.js
+++ b/app/assets/main/controllers/membership_choice_controller.js
@@ -25,6 +25,7 @@ export default class MembershipChoiceController extends Controller {
     if (!['free', 'single', 'multi'].includes(this.membership)) {
       return;
     }
+    this.membershipFieldTarget.value = this.membership;
     this.choiceTargets.forEach((target) => {
       if (target.value === this.membership) {
         MembershipChoiceController.activateChoice(target);
--- a/app/views/devise/registrations/_sign_up_form_without_subscription.html.erb
+++ b/app/views/devise/registrations/_sign_up_form_without_subscription.html.erb
@@ -119,7 +119,7 @@
 
       <button class="button button-cta w-full"
         data-target="registration-form.submitButton membership-choice.subscriptionInfo"
-        data-action="registration-form#submit"
+        data-action="registration-form#submitWithCardDetails"
         data-active-class="button button-cta w-full block"
         data-inactive-class="hidden"
       >
@@ -129,7 +129,7 @@
 
       <button class="button button-cta w-full hidden"
         data-target="registration-form.submitButton membership-choice.noSubscriptionInfo"
-        data-action="registration-form#submitWithoutCardDetails"
+        data-action="registration-form#submit"
         data-active-class="button button-cta block w-full"
         data-inactive-class="hidden"
       >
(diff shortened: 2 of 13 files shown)


app/views/devise/registrations/_sign_up_form_without_subscription.html.erb
```diff
@@ -117,10 +117,24 @@
...
+      </button>
+
+      <button class="button button-cta w-full hidden"
+        data-target="registration-form.submitButton membership-choice.noSubscriptionInfo"
+        data-action="registration-form#submitWithoutCardDetails"
+        data-active-class="button button-cta block w-full"
+        data-inactive-class="hidden"
+      >
+        <i class="fa fa-spinner fa-spin hidden" data-target="registration-form.loadingIndicator"></i>
+        <span><%= t 'views.registrations.create_account' %></span>
```
emmabckstrm, line 137: Two buttons instead of having a "handleChange" method in `registration-form` controller

spec/features/registrations_spec.rb
```diff
@@ -73,4 +73,34 @@
...
+    # Sign up page
+    find('input#free').click
+    find('#continue-to-payment').click
+    fill_in 'Email', with: 'test@example.com'
+    fill_in 'Password', with: 'password'
+    check 'I accept our Privacy policy'
+    click_button 'Create account'
+
+    # Wait for success page to render
+    find('.dashboard-show', wait: 20)
```
stefanlindbohm, line 104: Maybe add an expectation here to make the spec a bit clearer to read? Like the example above that says: `expect(page).to have_text 'Welcome to a climate neutral life'` (but updated with the text you get when signing up without a subscription)

stefanlindbohm:
Btw, there are some minor style changes in CodeClimate that would be nice to fix as well :). The duplication might be bigger so I'll let you decide on whether that's worth to fix though.

app/assets/main/controllers/membership_choice_controller.js
```diff
@@ -25,6 +25,7 @@ export default class MembershipChoiceController extends Controller {
     if (!['free', 'single', 'multi'].includes(this.membership)) {
       return;
     }
+    this.membershipFieldTarget.value = this.membership;
```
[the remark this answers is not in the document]
emmabckstrm: It's setting the value of the intended membership in the "registration form" form. Currently, the choose membership is its own form, and the registration form is a separate form on the second page... Thinking about it, I guess it could be refactored, but now I tried and apparently it's a whole different story to have forms within forms in Rails and I feel like that's a little too big of a project to take under now, especially with how much time I've already put into this form. @stefanlindbohm

emmabckstrm:
Summary:

Removed the A/B test for this feature, it's only behind the feature flag to be able to easily test it.
Also commented out the "sign up without subscription" tracking - will take care of that separately

PR #406, opened 2020-10-16
Remove Google tag manager tracking

emmabckstrm:
After discussions with @kallenilver we have decided to remove Google Tag Manager as it is not being used and not really relevant for our site. 

Regarding the tracking of local events or not - I have chosen to keep the scope on our current focus and added a card for handling local tracking later on https://trello.com/c/ZboMkrD4 - as it's not something that is being introduced now, rather an issue we have had since who knows.

@stefanlindbohm did you want some other sort of testing for this or is it fine to keep an eye on our tracking services in production?

### DIFF
--- a/app/views/application/_third_party_javascripts.html.erb
+++ b/app/views/application/_third_party_javascripts.html.erb
@@ -33,7 +33,6 @@
 <% if client_prefers_no_tracking? %>
   <script>
     function ga() {}
-    function gtag() {}
     function fbq() {}
   </script>
 <% else %>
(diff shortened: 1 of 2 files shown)


stefanlindbohm:
> @stefanlindbohm did you want some other sort of testing for this or is it fine to keep an eye on our tracking services in production?

This removal is great! No testing needed for this.

I was referring to adding a variant to the FB tracking that wouldn't send anything for non-production environments, like with GA we do `<% unless ENV['GOOGLE_ANALYTICS_SENDING'] == 'enabled' %>ga('set', 'sendHitTask', null);<% end %>`. And possibly in addition to that, enable logging of messages in dev, maybe something as simple as: `function fbq() { console.log(arguments) }` (unless there's a way to configure the FB SDK into a debug mode).

stefanlindbohm (approved):

emmabckstrm:
Cool, I'll add your great comment to the trello card! @stefanlindbohm

PR #408, opened 2020-10-19
Campaign / Know your footprint

emmabckstrm:
https://goclimateneutral-stagin-pr-408.herokuapp.com/know-your-carbon-footprint

There are still some content things that are left to fix, but other than that, I feel finished with the code parts of this and think it can be reviewed as is.

Todos
- [x] Waiting for content
- [x] Update translations with new content


app/assets/main/controllers/membership_choice_campaign_controller.js
```diff
@@ -0,0 +1,26 @@
+import { Controller } from 'stimulus';
+import { swapToActiveClassList, swapToInactiveClassList } from '../../util/swap_classes';
+
+export default class MembershipChoiceCampaignController extends Controller {
```
emmabckstrm, line 4: What we want to happen here is one small step more complicated than adding a general "ToggleController", thus I had to create this one. The name however can easily be changed 😅

app/controllers/users/registrations_controller.rb
```diff
@@ -16,7 +16,7 @@ def new
       @country_average = LifestyleFootprintAverage.find_by_country(@footprint.country)
 
       respond_to do |format|
-        format.html { render }
+        format.html { render params[:campaign].presence ? 'new_campaign' : 'new' }
```
emmabckstrm, line 19: Currently we only have one "campaign sign up" but this part could be extended if we need even more different campaign sign up pages

config/routes.rb
```diff
@@ -64,6 +64,10 @@
 
     resources :projects, path: 'climate-projects', only: [:index]
 
+    resource :know_your_footprint, controller: 'know_your_footprint', path: 'know-your-carbon-footprint', only: [:show]
+    get '/knowyourcarbonfootprint', to: redirect('know-your-carbon-footprint')
+    get '/KnowYourCarbonFootprint', to: redirect('know-your-carbon-footprint')
```
emmabckstrm, line 69: I'm a bit unsure about this part :) 
I think the first infrastructure is ok though, I'm mostly unsure about the extra urls - Tove wanted to have `/knowyourcarbonfootprint` and I added the capitalized one because why not. But not sure if there's a better way?

PR #411, opened 2020-10-26
Sum offsetting by persisted co2e

stefanlindbohm:
Commits should be pretty self-explanatory. I recommend viewing them one by one.


emmabckstrm (approved):
👏 👏 :shipit:

stefanlindbohm:
@kallenilver I'm merging this, but it would be great if you can take a look when you have time just to make sure you also agree with how we calculate sold offsetting.

kallenilver (commented):
I for one welcome our new sold offsetting calculations. 🍾

PR #413, opened 2020-10-28
Add referral codes to sign up page

stefanlindbohm:
Sign up page markup & JS is getting complicated, but this should work well for now.

# Todo

- [x] Make sure 3D Secure cards work (they will probably not have a payment intent for the first month and will instead have to use `SetupIntent`s to verify the card for future payments)
- [x] Make using referral codes case-insensitive
- [x] Localization
- [x] Rebase to master after #408 is merged and add referral code UI on the campaign page


stefanlindbohm:
This should be good now, I'll do another round of testing tomorrow with a fresher brain before merging in time for campaign start on Monday.

PR #419, opened 2020-10-30
Category visualization bug fix

emmabckstrm:


config/routes.rb
```diff
@@ -64,6 +64,10 @@
 
     resources :projects, path: 'climate-projects', only: [:index]
 
+    resource :know_your_footprint, controller: 'know_your_footprint', path: 'know-your-carbon-footprint', only: [:show]
```
emmabckstrm, line 67: dont need `controller: ` because same name

PR #423, opened 2020-11-03
Handle user that has no active subscription but has previously subscribed

emmabckstrm:
This PR updates of our evaluation of users with or without subscription, and what information on dashboard should be shown to either group. Additionally, it updates our admin dashboard to show more details on whether or not the new users have a subscription or not. The process can be followed commit per commit. The first commit is a migration which will be merged separately in https://github.com/goclimateneutral/goclimateneutral/pull/425

To show correct data in admin, we need to add value for `first_subscription_created_at`


app/controllers/admin/dashboard_controller.rb
```diff
@@ -30,6 +50,43 @@ def new_users
...
+
+    def new_users_without_subscription_with_referral_code
+      User.where('(first_subscription_created_at is null
+            OR date(first_subscription_created_at) > date(created_at))
+            AND referred_from_id is not null')
+          .order('date(created_at)')
+          .group('date(created_at)')
+          .count.transform_keys(&:to_s)
+    end
+
```
stefanlindbohm, line 89: I'd propose extracting this to a `UserStatistics` model or something like that, and also using SQL for the whole thing combined as this can be done relatively straightforwardly with one query.

app/models/subscription_manager.rb
```diff
@@ -39,6 +39,7 @@ def sign_up(plan, payment_method_id, referral_code = nil)
       else
         create_subscription(plan)
       end
+      user.update_first_subscription_created_at(Time.now)
```
stefanlindbohm, line 42: In cases where a user re-subscribes, this will overwrite a previous timestamp. I think the original should be kept with the reasoning that it's better to lose gaps in subscriptions rather than the original sign up date.

Also, I'd propose using `subscription.start_date` so that we use Stripe as the source of truth.

app/models/subscription_manager.rb
```diff
@@ -99,6 +100,7 @@ def create_subscription(plan, trial_end = nil)
       trial_end: trial_end&.to_i,
       expand: ['latest_invoice.payment_intent']
     )
+    user.update_first_subscription_created_at(Time.now) if user.first_subscription_created_at.nil?
```
stefanlindbohm, line 103: This should not be needed I think? Instead we should batch set these dates for existing subscriptions and always keep those values to the original sign up date.

app/models/user.rb
```diff
@@ -89,14 +93,22 @@ def update_from_stripe_subscription(stripe_subscription)
     save
   end
 
+  def update_first_subscription_created_at(date)
+    self.first_subscription_created_at = date
+    save
+  end
```
stefanlindbohm, line 99: Instead of creating this method, call points can use `user.update_attribute(:first_subscription_date, date)` which is a built in method with the same effect :).

app/models/user.rb
```diff
@@ -89,14 +93,22 @@ def update_from_stripe_subscription(stripe_subscription)
...
   def current_lifestyle_footprint
     lifestyle_footprints.order(:created_at).last
   end
 
-  def free_account?
-    return true unless stripe_customer&.subscriptions&.any?
+  def active_subscription?
+    stripe_customer.subscriptions&.first&.status == 'trialing' ||
+      stripe_customer.subscriptions&.first&.status == 'active'
+  end
```
stefanlindbohm, line 108: Shouldn't `.present?` be enough here? Ended subscriptions shouldn't be included in `.subscriptions` on a `Stripe::Customer` if I recall correctly.

app/views/admin/dashboard/index.html.erb
```diff
@@ -61,7 +141,7 @@
     <% end %>
   </tr>
   <tr>
-    <td>churend&nbsp;users</td>
+    <td>churned&nbsp;users</td>
```
stefanlindbohm, line 144: Yay, I've been bugged by this for a while but haven't gotten around to fixing it 🙈

app/views/dashboard/show.html.erb
```diff
@@ -217,13 +164,17 @@
 
   <section class="section-padding space-y-3 d:w-2/3">
     <h3 class="heading-lg mb-6">
-      <%=t 'you_have_lived_climate_neutral_for', count: @my_neutral_months %>
-      <%=t 'good_job' %>
+      <% if current_user.active_subscription? || current_user.subscription_months? %>
```
stefanlindbohm, line 167: One case this doesn't take into account now is for the very first page load after signing up with a subscription, where usually the first month hasn't yet been processed. Maybe we'll have to shuffle this around a bit and put some logic into the model that assumes at least 1 month as long as an active subscription exists?

stefanlindbohm (commented):
I'm very much liking these changes! I think I did catch a few things we should fix before merging, maybe let's pair on this tomorrow?

app/views/dashboard/show.html.erb
emmabckstrm: If I'm not misunderstanding, you are incorrect and this actually takes that scenario in consideration, hence the either an active subscription or having subscription months (in case a user has unsubscribed). It looks the same for everyone except users that neither have an active subscription nor any subscription months. @stefanlindbohm

PR #428, opened 2020-11-10
Show total users as member count

stefanlindbohm:
Since we now allow signing up for an account without starting a subscription and we have to goal of helping those users make an impact in other ways than offsetting too, we want to include those in the number of community members too.


kallenilver (commented):
🥳

PR #434, opened 2020-11-16
Update share section on dashboard

emmabckstrm:
Screenshot 2020-11-16 at 13 04 29
Screenshot 2020-11-16 at 13 04 49


stefanlindbohm (approved):
I like how this frames sharing as something to do that helps the climate, that makes it more of a call to action than just having a share button like everyone else :+1:. Good suff

PR #439, opened 2020-11-19
Cleanup assets

stefanlindbohm:
Notable changes:

- Drops jQuery and the JS portion of Bootstrap.
- Removes all old styling and related code.
- Changes Facebook sharing integration so that the Facebook SDK is only loaded after user interaction, so that we don't load it when not needed.
- Adds structure to `app/assets/main/controllers/`, putting page-specific JS controllers into directories based on the Rails controller they're used in (mirroring how templates are structured in `app/views/`). The idea being that we want to primarily use composable controllers, and they are left at first level, while moving distractions away for when you know you are changing a specific page. This is very much just a proposal, I'll add a discussion thread below.

Commit by commit is good too.


app/views/flight_footprints/new.html.erb
```diff
@@ -16,30 +16,30 @@
...
+            <div data-controller="flight-offsets--airport-field" data-search-endpoint="<%= airport_suggestions_path(format: :json) %>">
+              <div class="dropdown w-full" data-target="flight-offsets--airport-field.dropdownContainer">
+                <input type="text" required class="input w-full" data-target="flight-offsets--airport-field.searchField" data-action="input->flight-offsets--airport-field#search blur->flight-offsets--airport-field#select keydown->flight-offsets--airport-field#keyboardNavigate blur->flight-offsets--footprint-form#update">
+                <ul class="dropdown-menu cursor-pointer" data-target="flight-offsets--airport-field.suggestionsList flight-offsets--airport-field.dropdownTrigger" data-active-class="dropdown-menu cursor-pointer block">
                 </ul>
               </div>
-              <input type="hidden" name="destination_airport" data-target="airport-field.airportCodeField">
+              <input type="hidden" name="destination_airport" data-target="flight-offsets--airport-field.airportCodeField">
```
stefanlindbohm, line 42: Let's use this controller as an example for discussion for namespacing JS controllers. I've added namespaces for all controllers that are tied to a specific page. Probably best understood by looking at the file tree: https://github.com/goclimateneutral/goclimateneutral/tree/cleanup-assets/app/assets/main/controllers.

What are your thoughts on this, @emmabckstrm and @kallenilver?

app/views/flight_footprints/new.html.erb
```diff
@@ -16,30 +16,30 @@
...
+        <%= form_with url: url_for(action: :create, format: :json), html: { 'data-controller': 'flight-offsets--footprint-form' }, class: 'space-y-3' do |f| %>
           <div>
             <label class="font-bold">
               <%=t 'views.flight_footprints.new.from' %>
             </label>
-            <div data-controller="airport-field" data-search-endpoint="<%= airport_suggestions_path(format: :json) %>" >
-              <div class="dropdown w-full" data-target="airport-field.dropdownContainer">
-                <input type="text" required class="input w-full" data-target="airport-field.searchField" data-action="input->airport-field#search blur->airport-field#select keydown->airport-field#keyboardNavigate blur->flight-footprint-form#update">
-                <ul class="dropdown-menu cursor-pointer" data-target="airport-field.suggestionsList airport-field.dropdownTrigger" data-active-class="dropdown-menu cursor-pointer block">
+            <div data-controller="flight-offsets--airport-field" data-search-endpoint="<%= airport_suggestions_path(format: :json) %>" >
```
stefanlindbohm, line 24: `flight-offsets` should be `flight-footprints`

app/views/flight_footprints/new.html.erb
kallenilver: Sounds good, no objections from me.

kallenilver (approved):
🥳 

but I was hoping for smaller image file sizes as well :)

emmabckstrm (approved):
Well done! :clap:

stefanlindbohm:
> but I was hoping for smaller image file sizes as well :)

Hehe, sorry, I was too focused on the JS stuff. I'll take a look at the images in a separate PR :)

PR #440, opened 2020-11-20
New design gift cards

emmabckstrm:
The design is decently resilient against different wordings and sentence lengths. However, if the translated strings used in this template are way, way longer than their English counterparts, we might want to look into tweaking this even further.

As you can see in screenshot below, there are some rendering issues with the Inter font. It seems like the issue lies with Google fonts (https://github.com/google/fonts/issues/2586). The rendering issue is random, and sometimes it doesn't occur but most of the time it does. I see two solutions, either embedding the font in the gift card "template", or simply use the fall back font Arial. The former I have never done and might need some support in :)

Screenshot 2020-11-20 at 10 45 16


app/models/gift_card.rb
```diff
@@ -64,7 +64,7 @@ def update_from_payment_intent(payment_intent)
   end
 
   def order_id
-    "GCN-GIFT-#{id}"
+    "GC-GIFT-#{id}"
```
emmabckstrm, line 67: We're not GCN anymore :wink:

emmabckstrm: Also, I'm guessing this is just for presentation and won't affect any functionality?

stefanlindbohm (commented):
Code-wise looks good! Design-wise the only thing I'd like to propose is maybe to vertically align the three elements at the top? Looks a bit unstructured right now I think.

In terms of copy, I miss the specifics on how many tonnes the gift card is for and maybe a word or two on how "supporting a sustainable planet" ties in with the projects. Something to pair work on maybe?

kallenilver:
Regarding copy, yes, something that 6 months with average personal emissions of X ton for 6 months. Maybe a text at the bottom: "This gift card corresponds to the offsetting of the average emissions for one person living in USA for 6 months (10 tonnes of carbon dioxide emissions).".

emmabckstrm:
Screenshot 2020-11-23 at 15 57 08


Updated gift card, ping @stefanlindbohm @kallenilver 👋

kallenilver:
Looking good!

stefanlindbohm (approved):
Ship it!

PR #444, opened 2020-11-24
Update generated PDFs with Inter font and brand colours

emmabckstrm:
So, I realised yesterday that the font setup for gift cards that I thought work, didn't. After many tests I found out how to base64 encode the font and embed it directly in the template file. It seems to work, at least in review apps, which didn't work before. I went on and updated the font and colours for the rest of our PDFs: receipts, flight offset certicifactes, climate reports.

Screenshot 2020-11-25 at 15 40 06

Screenshot 2020-11-25 at 15 39 28


app/pdfs/climate_report_pdf.rb
```diff
@@ -40,10 +47,7 @@ def render
         }
       ),
       footer: {
-        right: 'Sida [page] av [topage]',
```
kallenilver, line 43: So page numbers are removed here?

kallenilver (approved):
Wowz, looking good!

app/pdfs/climate_report_pdf.rb
emmabckstrm: Page numbers will be rendered, and they will show the same as before, it's just that I had to extract the footer to its own file in order to embed the font. So when the PDF is generated, the same information is visible as before :) @kallenilver

stefanlindbohm:
Very nice change!!

Can I request that the font embed markup be extracted into a partial? Because I can't even seem to view the diffs here on Github with the font embedded in the templates, so loading those files into an editor seems bound to be a bit sluggish. Also that way we don't have the same data committed in our repo into multiple files :).

emmabckstrm:
@stefanlindbohm I can try! I tried one way, and it didn't work but perhaps there's one more thing I can tinker with. Maybe don't get your hopes up too far though :) 
and honestly it's a bit of a pain with the huge file but it's manageable for me at least. I have opened way worse text files, hehe

emmabckstrm:
@stefanlindbohm for reference, I'll release this and look into file management in a separate PR.

PR #448, opened 2020-11-26
Display shopping separately from public emissions in footprint charts

stefanlindbohm:
See discussion on Slack.

Screen Shot 2020-11-26 at 10 23 50
Screen Shot 2020-11-26 at 10 16 00

### DIFF
--- a/app/views/shared/_chart_categories.html.erb
+++ b/app/views/shared/_chart_categories.html.erb
@@ -1,6 +1,6 @@
 <div class="relative pb-1">
   <div class="space-y-6">
-    <% max_value_categories = [@footprint.housing.tonnes.round(1), @footprint.food.tonnes.round(1), @footprint.car.tonnes.round(1), @footprint.flights.tonnes.round(1), (@footprint.public.tonnes.round(1) + @footprint.consumption.tonnes.round(1))].max %>
+    <% max_value_categories = [@footprint.housing.tonnes.round(1), @footprint.food.tonnes.round(1), @footprint.car.tonnes.round(1), @footprint.flights.tonnes.round(1), @footprint.public.tonnes.round(1), @footprint.consumption.tonnes.round(1)].max %>
     <% [
         {
           amount: @footprint.housing,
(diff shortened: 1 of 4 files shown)


app/views/shared/_chart_categories.html.erb
```diff
@@ -27,12 +27,18 @@
           title: t('dashboard.footprint.category.flights'),
           'bg-color': 'bg-blue-accent'
         },
+        {
```
emmabckstrm, line 30: Shouldn't `(@footprint.public.tonnes.round(1) + @footprint.consumption.tonnes.round(1))` be separated here as we are separating the values?

app/views/shared/_chart_categories.html.erb
```diff
@@ -27,12 +27,18 @@
...
+          'bg-color': 'bg-pink-accent'
+        },
         {
           key: 'public',
-          amount: @footprint.public + @footprint.consumption,
+          amount: @footprint.public,
           icon: 'fa-university',
           title: t('dashboard.footprint.category.public'),
-          'bg-color': 'bg-pink-accent'
+          'bg-color': 'bg-gray-accent'
```
emmabckstrm, line 41: Not really relevant to this specific PR but it makes me think that perhaps we need one more colourful brand colour :D

app/views/shared/_chart_categories.html.erb
emmabckstrm: At row 3 as I couldn't really comment on that specific row as it's not edited

app/views/shared/_chart_categories.html.erb
stefanlindbohm: I actually think the gray fit quite well given it's the category you can't change directly. But yeah :D

app/views/shared/_chart_categories.html.erb
stefanlindbohm: Oh yeah, missed that. Thanks!

PR #449, opened 2020-11-26
Upgrade to Font Awesome 5

stefanlindbohm:
Icons we used for footprint charts are better visually balanced with this. And we have access to more icons should we need them.

## Todo

- [x] Update icons added by #448 after that PR is merged.

## Screenshots

Before:
Screen Shot 2020-11-26 at 11 53 51

After:
Screen Shot 2020-11-26 at 11 53 00

### DIFF
--- a/app/assets/bundles/main.js
+++ b/app/assets/bundles/main.js
@@ -31,7 +31,10 @@ application.load(definitionsFromContext(
 ));
 
 /* Font Awesome */
-import 'font-awesome/scss/font-awesome.scss';
+import '@fortawesome/fontawesome-free/scss/fontawesome.scss';
+import '@fortawesome/fontawesome-free/scss/solid.scss';
+import '@fortawesome/fontawesome-free/scss/regular.scss';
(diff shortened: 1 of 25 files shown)


emmabckstrm:
Nice! I agree it's looking really good! I see that we are using different  font awesome classes. I assume `fab` are specific for brand icons? But the others, `fas` and `far` - should we add some small documentation on when we are using which type?

stefanlindbohm:
> Nice! I agree it's looking really good! I see that we are using different font awesome classes. I assume `fab` are specific for brand icons? But the others, `fas` and `far` - should we add some small documentation on when we are using which type?

Yep, b for brand icons. `fas` for solid (previous default) and `far` for "regular" which is outlines. The change with Font Awesome 5 is that there are more outline ("regular") variants available, and the namespace separation. See https://fontawesome.com/how-to-use/on-the-web/referencing-icons/basic-use.

Documenting this sounds good to me. I'm thinking of where the right place is, maybe in the design kit Sketch file? Since this is really part of the design process?

PR #451, opened 2020-11-27
Add A/B test for removing signup without subscription

emmabckstrm:
I simply just hid the "free" alternative in order to change as little as possible and have the test be as clean as possible.

Screenshot 2020-11-27 at 12 33 42 

Screenshot 2020-11-27 at 12 33 25


kallenilver (approved):
Yey for this and for us learning more!

stefanlindbohm (commented):
Good stuff!

PR #453, opened 2020-11-28
Make the certificate a bit more piff

kallenilver:
I think the certificate gets a bit more piffy with a background image. Feels important with piffyness since it will be hanged on walls etc.
Before:
Skärmavbild 2020-11-28 kl  17 07 17
After:
Skärmavbild 2020-11-28 kl  17 07 46


emmabckstrm (approved):
Very nice! 👍

stefanlindbohm (approved):
Dandy dandy!

PR #455, opened 2020-12-01
Added new flags for new regions and compressed all flags

emmabckstrm:
I hope I got all the new flags!


stefanlindbohm:
Integrated in #456

PR #457, opened 2020-12-11
Upgrade and cleanup frontend

stefanlindbohm:
A bunch of small-ish things we have talked about and I've had itchy fingers about for a while. Some might be points of discussion. Commit-by-commit is probably best for reviews.


emmabckstrm (approved):
feels good! great work! 

feelsgood

PR #458, opened 2020-12-14
Sign up for newsletter

emmabckstrm:
I did some changes to make this work well and look good

- Added `sm` versions of our button and input CSS components
- New table (will be merged in separate PR)
- Small updates to design of footer to keep it a little more visually balanced
- Some changes in specs that could't distinguish the "email" field the were supposed to use when logging in vs the new I added (don't know why it had issues, the only name that was similar was the placeholder in the new input field 🤷‍♀️ )

Screenshot 2020-12-10 at 11 18 28

### DIFF
--- a/app/controllers/newsletter_subscribers_controller.rb
+++ b/app/controllers/newsletter_subscribers_controller.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+class NewsletterSubscribersController < ApplicationController
+  def create
+    params.require(:newsletter_email)
+    params.permit(:logged_in_user_id, :region)
+    subscriber = NewsletterSubscriber.new(
+      email: params[:newsletter_email],
+      logged_in_user_id: params[:logged_in_user_id],
+      region: params[:region]
+    )
+
+    if subscriber.save
+      head :ok
+    else
+      render(
+        status: :bad_request,
+        json: { error: { message: 'email' } }
+      )
+    end
+  end
+end
--- a/app/views/shared/_footer.html.erb
+++ b/app/views/shared/_footer.html.erb
@@ -5,10 +5,12 @@
 
   <section class="bg-primary text-white text-sm p-4 py-6 t:px-8 t:py-12 space-y-4">
     <div class="flex flex-row flex-wrap -m-4">
-      <div class="w-full d:w-2/5 flex-grow p-4">
-        <h3 class="font-bold">GoClimate</h3>
-        <p class="max-w-lg"><%= t('views.shared.footer.short_intro') %></p>
-        <ul class="list-none pt-4 flex flex-row space-x-4">
+      <div class="w-full d:w-2/5 flex-grow p-4 space-y-4">
+        <div>
+          <h3 class="font-bold">GoClimate</h3>
+          <p class="max-w-lg"><%= t('views.shared.footer.short_intro') %></p>
+        </div>
+        <ul class="list-none flex flex-row space-x-4">
           <li>
             <%= link_to 'https://www.instagram.com/goclimate/', target: "_blank", rel: "noopener" do %>
               <i class="fab fa-2x fa-instagram" aria-hidden="true"></i>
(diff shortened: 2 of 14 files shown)


db/migrate/20201211094835_create_newsletter_subscribers.rb
```diff
@@ -0,0 +1,12 @@
+class CreateNewsletterSubscribers < ActiveRecord::Migration[6.0]
+  def change
+    create_table :newsletter_subscribers do |t|
+      t.timestamps
+      t.string :email, null: false
```
emmabckstrm, line 5: Hmm, should this perhaps be unique? So we don't get duplicates/get spammed?

config/routes.rb
```diff
@@ -134,6 +134,8 @@
       end
     end
 
+    resources :newsletter_subscribers
```
emmabckstrm, line 137: Hmm, should this be restricted to only the action(s) actually being used?

app/views/shared/_footer.html.erb
```diff
@@ -28,6 +30,36 @@
             <% end %>
           </li>
         </ul>
+
+        <div>
+          <%= form_with url: newsletter_subscribers_path, html: { 'data-controller': 'newsletter-subscription' } do |f| %>
```
stefanlindbohm, line 35: This would be a great place to use `form_with url: NewsletterSubscriber.new(logged_in_as_user_id: current_user&.id)` which would ensure we only use attributes that exist on the model, and also simplify the controller code to this:

```ruby
subscriber = NewsletterSubscriber.new(params.require(:newsletter_subscriber).permit(:email, :logged_in_as_user_id, :country)
```

app/controllers/newsletter_subscribers_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+class NewsletterSubscribersController < ApplicationController
+  def create
+    subscriber = NewsletterSubscriber.new(
+      email: params[:newsletter_email],
+      logged_in_as_user_id: params[:logged_in_as_user_id],
+      country: params[:region]
+    )
+
+    if subscriber.valid?
+      subscriber.save
```
stefanlindbohm, line 12: This can be simplified to `if subscriber.save`

app/controllers/newsletter_subscribers_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+      render json: {
+        status: 200
+      }
```
stefanlindbohm, line 15: Opinion here, but since the HTTP status is already 200 OK, I would lean towards just an empty response. As in `head :ok` instead of `render json: ...`. The client can check for the HTTP status code directly (which is actually slightly so simpler with `fetch` in JS I think).

config/routes.rb
stefanlindbohm: Yep indeed!

db/migrate/20201211094835_create_newsletter_subscribers.rb
stefanlindbohm: The one argument I can think of to now make it unique is that someone might subscribe again after a long period of time, and we might want to or have to periodically purge old addresses? But probably not?

If you do add validation, do it both here and in ActiveRecord though. Or just in ActiveRecord. If we do it only in the DB, ActiveRecord crashes and behaves weirdly unnecessarily. DB constraints are basically for protecting against race conditions.

db/migrate/20201211094835_create_newsletter_subscribers.rb
```diff
@@ -0,0 +1,12 @@
...
+      t.string :country
```
stefanlindbohm, line 6: Since we're actually storing the current region here, I'd vote for renaming this field to `region` to make that clear.

db/migrate/20201211094835_create_newsletter_subscribers.rb
```diff
@@ -0,0 +1,12 @@
...
+      t.bigint :logged_in_as_user_id
```
stefanlindbohm, line 7: Just very nitpicky, but I'd vote for `logged_in_user_id` here, without the `as`

stefanlindbohm (commented):
Nicely put together! I consider my comments technicalities, but that's my thing I guess :).

PR #460, opened 2020-12-14
Flexible reports in admin

stefanlindbohm:
Just opening this to make this work visible. It's going to change a lot.


stefanlindbohm:
I've experimented different ways and am currently thinking that Mode (https://mode.com) will be a better solution. Pretty much the same of allowing arbitrary SQL to be presented in charts but without code we have to maintain.

PR #461, opened 2020-12-18
Update your footprint

emmabckstrm:
## Subscription settings
Screenshot 2020-12-22 at 15 51 53

## Dashboard
Subscriber
Screenshot 2020-12-22 at 15 52 06

Non-subscriber
Screenshot 2020-12-22 at 15 53 29

## Previous footprints
Screenshot 2020-12-22 at 15 52 12

## Showing a footprint
For person who is the owner of the footprint
Screenshot 2020-12-22 at 15 52 41

For anyone else that is not the owner of the footprint
Screenshot 2020-12-23 at 10 43 48

Deleting a footprint
Screenshot 2020-12-22 at 15 53 44


app/controllers/dashboard_controller.rb
```diff
@@ -20,10 +21,27 @@ def show
...
     @projects = Project.all.order(created_at: :desc).limit(5)
+
+    @current_plan_price = @manager.subscription&.plan&.monthly_amount
+    if @footprint.present?
+      @full_footprint_price = SubscriptionManager.price_for_footprint(
+        current_user.current_lifestyle_footprint&.total,
+        customer_currency
+      )
+    end
+    @total_number_of_footprints = LifestyleFootprint.where(user_id: current_user.id).length
```
stefanlindbohm, line 32: Using `.length` will trigger the ActiveRecord result to retrieve all records and convert itself into a collection. For performance, we should use `.count` instead which makes ActiveRecord instead run an SQL `COUNT` aggregation query in cases like this where that's all we're interested in.

app/views/dashboard/show.html.erb
```diff
@@ -65,13 +65,38 @@
                 <%= t("dashboard.footprint.your_climate_footprint", **footprint_text_options) %>
               <% end %>
             </p>
+            <% if @current_plan_price && @full_footprint_price %>
+              <p class="hidden t:block">
+                <%= t("views.dashboard.footprint.subscription_coverage",
+                  subscription_price: @current_plan_price.to_s(precision: :auto),
+                  footprint_coverage: ((@current_plan_price.amount/@full_footprint_price.amount * 100).round))
```
stefanlindbohm, line 72: I'd vote for doing this calculation in the controller straight and use a variable named `@footprint_coverage` instead, to make it clear that this is what we're interested in since `@full_footprint_price` is never used in any other way.

app/views/dashboard/show.html.erb
```diff
@@ -65,13 +65,38 @@
...
+        <div class="inline-block">
+          <%= link_to t('views.dashboard.footprint.previous_footprints'), user_lifestyle_footprints_path, class: 'link whitespace-nowrap' %>
+        </div>
+      <% end %>
+      <div class="inline-block">
+        <%= link_to (current_user.active_subscription? ? t('views.dashboard.footprint.update_subscription') : t('views.dashboard.footprint.subscribe')),
+          user_subscription_path,
+          class: 'link whitespace-nowrap'
+        %>
+      </div>
```
stefanlindbohm, line 98: Maybe you've experimented with this already, in which case you can ignore me here, but would this look a little more uniform if all of these looked like buttons? Something about links next to buttons is throwing me off a bit.

app/controllers/lifestyle_footprints_controller.rb
```diff
@@ -22,11 +24,33 @@ def create
...
+    else
+      redirect_to new_registration_path(:user, **query_params)
+    end
+  end
+
+  def show
+    render_not_found && return unless @footprint.present?
+
+    # Some old users don't have LifestyleFootprints
+    @country_average = LifestyleFootprintAverage.find_by_country(@footprint.country) if @footprint.present?
```
stefanlindbohm, line 38: The comment and conditional here are redundant with the guard clause just above, right?

app/controllers/lifestyle_footprints_controller.rb
```diff
@@ -22,11 +24,33 @@ def create
...
+    @total_number_of_footprints = LifestyleFootprint.where(user_id: current_user.id).length if current_user.present?
```
stefanlindbohm, line 39: Another `.length` -> `.count`

app/controllers/lifestyle_footprints_controller.rb
```diff
@@ -22,11 +24,33 @@ def create
...
+  def destroy
+    @footprint.destroy
+    flash[:notice] = 'Your carbon footprint calculation was deleted sucessfully.'
+    redirect_to dashboard_path
   end
 
   private
 
+  def set_footprint
+    @footprint = LifestyleFootprint.where(key: params[:id]).first
```
stefanlindbohm, line 51: This can be simplified as `LifestyleFootprint.find_by_key(params[:id])`. And adding a bang (`.find_by_key!`) will actually remove the need for the guard clause in the actions themselves, as that will raise an error that makes Rails respond with a 404 page automatically when nothing is found.

app/views/lifestyle_footprints/show.html.erb
```diff
@@ -0,0 +1,119 @@
...
+                <% end %>
+              <% end %>
+            </p>
+          <% end %>
+        </div>
+        <div class="t:w-1/2">
+          <%= render "shared/chart_categories" %> 
+        </div>
+      </div>
+    </div>
```
stefanlindbohm, line 53: Some or all of this is the same as on dashboard, right? A good time for a partial for this information? To me, this page showing previous results isn't as important so the cost of maintaining another component doesn't seem worth it to do much customization here anyway.

app/controllers/users/lifestyle_footprints_controller.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Users
+  class LifestyleFootprintsController < ApplicationController
```
stefanlindbohm, line 4: Why a separate controller for this? The `index` method makes sense as part of the no-prefix `LifestyleFootprintsController`, doesn't it?

app/views/users/subscriptions/show.html.erb
```diff
@@ -32,10 +32,18 @@
                 <% @available_plans.each do |i| %>
                   <option value="<%= i.amount %>" <%= i == @current_plan_price ? "selected" : "" %> >
                     <%= i.to_s(precision: :auto) %>
+                    (<%= i == @full_footprint_price ?
+                        current_user.current_lifestyle_footprint.total.tonnes.round(1) :
+                        (current_user.current_lifestyle_footprint.total.tonnes * (i.amount/@full_footprint_price.amount)).round(1)
+                      %>
```
stefanlindbohm, line 38: I'm confused by this logic, although I understand how you ended up with it now that I realize what it's doing. May I propose we use `GreenhouseGases.from_consumer_price` instead? That's after all the same logic that we use when a payment comes in and we define how many kg's it corresponds to. (Yes, I consider this deprecated, but the solution to me is that each subscription plan as a pre-defined amount of CO2e per month. Until then the deprecated behavior is what we actually do use. (Maybe deprecated isn't actually the right word since we have no replacement, but that's a discussion for elsewhere I guess.))

config/locales/views/dashboard.en.yml
```diff
@@ -1,6 +1,11 @@
 en:
   views:
     dashboard:
+      footprint:
+        subscription_coverage: Your current monthly offsetting is %{subscription_price} which corresponds to %{footprint_coverage}% of your yearly carbon footprint.
+        previous_footprints: See previous footprints
+        update_subscription: Update subscription
+        subscribe: Balance your footprint
```
stefanlindbohm, line 8: I am really not sure about using the word "balance" as a CTA without context here. Have you considered alternatives? It seems fluffy to me. Not sure what I would propose instead. Maybe straight-up "Start a subscription based on your footprint"? Long-term I feel this should warrant a dedicated section on this page, describing our subscriptions similarly to the sign up page, but given that we probably shouldn't scope creep this PR to include that I'm OK with whatever you end up deciding on.

config/routes.rb
```diff
@@ -60,7 +61,8 @@
     resource :privacy_policy, controller: 'about/privacy_policy', path: 'privacy-policy', only: [:show]
     resource :travel_calculator, controller: 'travel_calculator', path: 'travel-calculator', only: [:show]
 
-    resources :lifestyle_footprints, path: 'calculator', only: [:new, :create], path_names: { new: '' }
+    resources :lifestyle_footprints, path: 'calculator', only: [:new, :create, :destroy], path_names: { new: '' }
+    get 'calculator/result/:id', to: 'lifestyle_footprints#show', as: 'lifestyle_footprints_result'
```
stefanlindbohm, line 65: Maybe this is me being deep in la la Rails land, but I would prefer this to clarify that this is really the same resource, but with different paths:

```suggestion
    resources :lifestyle_footprints, path: 'calculator', only: [:new, :create], path_names: { new: '' } do
      collection do
        resources :lifestyle_footprints, path: 'result', only: [:show, :destroy]
      end
    end
```

(This also changes the path helper from `lifestyle_footprint_result` to `lifestyle_footprint` the same way that standard `show` actions get the resource in singular as their names.)

tailwind.config.js
```diff
@@ -27,6 +27,7 @@ module.exports = {
       black: '#000',
       white: '#fff',
       primary: '#1C4637',
+      danger: '#CC412E',
```
stefanlindbohm, line 30: I'm not for or against, really, but didn't we say we shouldn't use semantic color names? Maybe I've made the case for (or against) this in the past, but I can't remember now 😅.

stefanlindbohm (approved):
Nicely put together!

My comments are all Rails technicalities or nitpicks, so consider it all as green on our feedback scale :).

config/routes.rb
stefanlindbohm: If you agree to move the index action here as well, this is what I'd suggest:

```suggestion
    resources :lifestyle_footprints, path: 'calculator', only: [:new, :create], path_names: { new: '' } do
      collection do
        resources :lifestyle_footprints, path: 'results', only: [:index, :show, :destroy]
      end
    end
```

(Add the plural `s` to `results` and the action itself.)

app/views/users/subscriptions/show.html.erb
emmabckstrm: I can't seem to wrap my head around this -  I wrote to you on slack about it

emmabckstrm (commented):
Great feedback @stefanlindbohm - I have addressed most of it in comments, if you are interested

PR #465, opened 2020-12-30
Add shopping question to lifestyle calculator

stefanlindbohm:
Pretty much boils down to this screenshot, see discussion on Slack.

Screen Shot 2020-12-30 at 13 46 01

## Todo

- [ ] Translate new strings into Swedish

### DIFF
--- a/spec/factories/lifestyle_calculator.rb
+++ b/spec/factories/lifestyle_calculator.rb
@@ -39,6 +39,12 @@ FactoryBot.define do
         { key: 'omnivore', formula: '2000' }
       ]
     end
+    shopping_options do
+      [
+        { key: 'zerowaste', formula: '1000' },
+        { key: 'average', formula: '1500' }
+      ]
+    end
     car_type_options do
       [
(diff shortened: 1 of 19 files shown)


spec/factories/lifestyle_calculator.rb
```diff
@@ -39,6 +39,12 @@
         { key: 'omnivore', formula: '2000' }
       ]
     end
+    shopping_options do
+      [
+        { key: 'zerowaste', formula: '1000' },
+        { key: 'average', formula: '1500' }
+      ]
```
emmabckstrm, line 46: Maybe it's my brain that is slow but shouldn't it be a third on here?

emmabckstrm (approved):
Nice! Just had one question

spec/factories/lifestyle_calculator.rb
stefanlindbohm: This is just the factory used in testing. Actual data lives in the production DB :).

PR #467, opened 2021-01-21
New design business page

emmabckstrm:
New design - copy is discussed with Tove. Mobile view is decent.

Screenshot 2021-01-25 at 11 11 28
Screenshot 2021-01-25 at 11 11 35
Screenshot 2021-01-25 at 11 11 42
Screenshot 2021-01-25 at 11 11 47


stefanlindbohm (approved):
Very nice change!

I'd keep the logos directly after the first section if it were up to me. Reasoning that most SaaS websites do this and I've actually found myself looking for that kind of peer validation while researching products lately. It's a good quick gauge for whether something is worth looking further into if it seems they have companies you know about as customers.

emmabckstrm:
@stefanlindbohm great point! I agree! We had that before, as well, so I'll make that change and deploy!

PR #470, opened 2021-01-26
Temporary admin-tool to batch calculate emissions from flights

kallenilver:
Most likely to be removed when we build a better tool for our customers to use. But hopefully useful until then.


stefanlindbohm (approved):
Impressively quick and dirty :). It's in isolation so I'll say it's fine :+1:. Maybe add some comment in the code stating the intended lifespan of it?

PR #471, opened 2021-02-02
Remove AB test for signup without subscription

emmabckstrm:


stefanlindbohm (approved):
Yay! This should in theory mean we get double the amount of signups for free accounts 🥳

PR #472, opened 2021-02-03
Add React

stefanlindbohm:


stefanlindbohm:
@emmabckstrm Maybe we can create some separation so that React isn't loaded when not in use and merge this PR? That way we have it available when we're ready to start using it.

emmabckstrm:
@stefanlindbohm that sounds like a great idea!

PR #473, opened 2021-02-11
Cleanup fix for region being dropped in signup since e6f5ca1

stefanlindbohm:
Cleans up from quick fix in 851756f.

A few learnings:

- Passing a single object to a path helper that doesn't take an ID
  replaces the implicit use of params for that request (in this case
  causing the disappearance of the region param in the path)
- With routes for the same resources defined twice, form_with seems
  confused about which of them to use, even when each action is only
  present in either of the two routes. Specifying both model and url
  solves this.

### DIFF
--- a/app/controllers/lifestyle_footprints_controller.rb
+++ b/app/controllers/lifestyle_footprints_controller.rb
@@ -14,26 +14,21 @@ class LifestyleFootprintsController < ApplicationController
 
     render_not_found && return unless @calculator.present?
 
-    @footprint = LifestyleFootprint.new(lifestyle_calculator: @calculator)
+    @footprint = LifestyleFootprint.new(lifestyle_calculator: @calculator, country: params[:country])
   end
 
   def create
-    campaign_param = params[:campaign]
-
     @footprint = LifestyleFootprint.new(footprint_params)
     @footprint.update_from_lifestyle_calculator
     @footprint.user = current_user
     @footprint.save!
 
-    query_params = {
-      lifestyle_footprint: @footprint
-    }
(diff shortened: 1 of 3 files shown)


emmabckstrm (changes requested):
Nice work! This really cleans everything up as it mostly should be! However, it introduces the signup url having `campaign=` if there is no value for the campaign and I feel that would look unprofessional (I tested it locally). My suggestion is to bring back the 

```
    query_params = {
      lifestyle_footprint: @footprint
    }
    query_params[:campaign] = campaign_param unless campaign_param.blank?
 ```
 
 part, if there is no one liner way of altering as it is now

kallenilver:
May I also add a test to the wishlist so that we can make sure this does not happen again?

stefanlindbohm:
@emmabckstrm 

> Nice work! This really cleans everything up as it mostly should be! However, it introduces the signup url having `campaign=` if there is no value for the campaign and I feel that would look unprofessional (I tested it locally). My suggestion is to bring back the
> 
> ```
>     query_params = {
>       lifestyle_footprint: @footprint
>     }
>     query_params[:campaign] = campaign_param unless campaign_param.blank?
> ```
> 
> part, if there is no one liner way of altering as it is now

Haha, I thought I did just that but messed up before I made the commit. Fixed now (using `.presence` and no conditionals nor the stuff I did that didn't do anything at all).

@kallenilver 

> May I also add a test to the wishlist so that we can make sure this does not happen again?

Added to the integration test!

emmabckstrm (approved):
⭐ ⭐ ⭐

kallenilver (approved):
🥳

PR #476, opened 2021-03-08
Admin views for creating orginaztions and climate reports

emmabckstrm:
## NOTE!

This PR changes the order of the migrations, meaning that before running this migration, production db needs to rollback two steps.

### DIFF
--- a/app/controllers/admin/climate_reports_controller.rb
+++ b/app/controllers/admin/climate_reports_controller.rb
@@ -1,0 +1,63 @@
+# frozen_string_literal: true
+
+module Admin
+  class ClimateReportsController < AdminController
+    before_action :set_report, only: [:show, :edit, :update, :destroy]
+    before_action :set_organizations, only: [:new, :edit]
+
+    def index
+      @reports = ClimateReports::Report.all.order(title: :asc)
+    end
+
+    def show
+    end
+
+    def new
+      @report = ClimateReports::Report.new
+    end
+
+    def edit
+    end
+
+    def create
+      @report = ClimateReports::Report.new(report_params)
+
+      if @report.save
+        redirect_to [:admin, @report], notice: 'Report was successfully created.'
+      else
+        render :new
+      end
+    end
+
+    def update
+      if @report.update(report_params)
+        redirect_to [:admin, @report], notice: 'Report was successfully updated.'
+      else
+        render :edit
+      end
+    end
+
+    def destroy
+      @report.destroy
+
+      redirect_to admin_climate_reports_url, notice: 'Report was successfully destroyed.'
+    end
+
+    private
+
+    def set_report
+      @report = ClimateReports::Report.find(params[:id])
+    end
+
+    def set_organizations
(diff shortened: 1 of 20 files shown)


app/controllers/admin/climate_reports_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+    end
+
+    def create
+      @report = ClimateReports::Report.new(report_params)
+
+      if @report.save
+        redirect_to [:admin, @report], notice: 'Report was successfully created.'
+      else
+        render :new
+      end
```
stefanlindbohm, line 29: I would do this as a guard clause for the non-success path:

```ruby
unless @report.save
  render :new
  return
end

redirect_to [:admin, @report], notice: 'Report was successfully created.'
```

app/controllers/admin/climate_reports_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+    end
+
+    def update
+      if @report.update(report_params)
+        redirect_to [:admin, @report], notice: 'Report was successfully updated.'
+      else
+        render :edit
+      end
```
stefanlindbohm, line 37: Same as above.

app/controllers/admin/climate_reports_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    def set_organizations
+      @organizations = Organization.all.order(name: :asc).map { |o| [o.name, o.id] }
+    end
+
+    def report_params
+      params.require(:climate_report).permit(:title, :start_date, :end_date, :organization).tap do |p|
+        p[:reporting_period] = p[:start_date]..p[:end_date]
+        p[:organization_id] = p[:organization]
+      end.except(:start_date, :end_date, :organization)
```
stefanlindbohm, line 60: Nice solution 😏

stefanlindbohm: ...but why isn't `organization` just named `organization_id` in all places?

app/controllers/admin/organizations_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+
+    def create
+      @organization = Organization.new(organization_params)
+
+      if @organization.save
+        redirect_to [:admin, @organization], notice: 'Organization was successfully created.'
+      else
+        render :new
+      end
```
stefanlindbohm, line 28: I'd prefer a guard clause here too

app/controllers/admin/organizations_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+
+    def update
+      if @organization.update(organization_params)
+        redirect_to [:admin, @organization], notice: 'Organization was successfully updated.'
+      else
+        render :edit
+      end
```
stefanlindbohm, line 36: Once more.

emmabckstrm:
@stefanlindbohm 's comments are addressed in https://github.com/goclimate/goclimate/commit/645be389ca6e9a7e662ce6f1ca776a0f5ba9fe89

PR #479, opened 2021-03-19
Functionality for destroying business calculators, fields and categories

emmabckstrm:


emmabckstrm:
[Possibility to remove fields and categories from business calculators](https://trello.com/c/mR0EYQb5/585-possibility-to-remove-fields-and-categories-from-business-calculators)

PR #481, opened 2021-03-25
Add preview of data collection form in admin

emmabckstrm:


emmabckstrm:
[Preview User form from step 1](https://trello.com/c/59ucV2Sf/569-preview-user-form-from-step-1)

PR #486, opened 2021-04-19
Add data request email and send when creating a data request

emmabckstrm:


app/mailers/data_request_mailer.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class DataRequestMailer < ApplicationMailer
+  default asm: { group_id: SENDGRID_ASM_GROUP_IDS[:welcome] }
```
emmabckstrm, line 4: This needs to be properly aligned

app/mailers/data_request_mailer.rb
```diff
@@ -0,0 +1,17 @@
...
+  helper :application
```
emmabckstrm, line 5: I'm not sure if this is needed.

config/locales/views/mailers/data_requests.sv.yml
```diff
@@ -0,0 +1,5 @@
+sv:
+  mailers:
+    data_requests:
+      subject: Du har blivit ombedd att dela med dig av data
```
emmabckstrm, line 4: Might be the worst placeholder copy I've ever written

PR #487, opened 2021-04-20
Collective efficacy and impact test in sign up flow

kallenilver:


app/views/devise/registrations/_chart_world_comparison.html.erb
```diff
@@ -40,3 +40,8 @@
     <%= t("views.registrations.your_climate_footprint", **footprint_text_options) %>
   <% end %>
 </p>
+<p>
+  <% if experiment_active?(:impact) %>
+    <b><%= t("views.registrations.impact_test") %></b>
```
emmabckstrm, line 45: Firstly - I would move the `<p>` tag into the if clause - otherwise you'll have an empty `<p>` tag when the condition is not met - or is there a reason for this?

Secondly - and I'm not sure about this one - do the `<b>` tags provide boldness or do you need to add a class here? If adding class, I would remove the `<b>` tag and put the class on the `<p>` tag

README.md
```diff
@@ -124,7 +124,7 @@ code `eo` in this project). This is done by running: `bin/rails i18n:normalize_c
 We have a custom set up for A/B testing. Setting up an A/B test is done in `experiments.yml`. When you have added the feature key name and frequency in `experiments.yml`, you are ready to use it in templates.
 
 ```
-  <% if @experiment_active?(:my_feature) %>
+  <% if experiment_active?(:my_feature) %>
```
emmabckstrm, line 127: 😍 😍

emmabckstrm (approved):
I just had a minor thing, otherwise it looks good!

app/views/devise/registrations/_chart_world_comparison.html.erb
kallenilver: Great, thank you! Fixed it!

PR #488, opened 2021-04-20
Removing input type change when changing unit type when creating new reported data

emmabckstrm:
This is due to Javascript limitations and the fact that the value of the input field is cleared when changing the input type, if the value is not allowed by the new type.


emmabckstrm:
[Input value is lost when changing unit](https://trello.com/c/67TPmdoA/599-input-value-is-lost-when-changing-unit)

PR #490, opened 2021-04-23
Better structure for viewing data collection result in admin view

emmabckstrm:


emmabckstrm:
[Second iteration of admin view to show reported data](https://trello.com/c/qYkVnaC1/602-second-iteration-of-admin-view-to-show-reported-data)

PR #491, opened 2021-04-23
A/B-test for partners section on index page

kallenilver:
So the hypothesis is that showing these partners will give us a little boost in trust which will increase conversion. Showing badges and brands seems to work for e-commerce, but these logos might not be as well known as many others. Lets let data speak.

Skärmavbild 2021-04-23 kl  21 42 21


emmabckstrm (approved):
:clap:

PR #492, opened 2021-04-26
Data requests with custom email content

emmabckstrm:


emmabckstrm:
[Automatic emails not always good](https://trello.com/c/7HemdMDy/611-automatic-emails-not-always-good)

app/mailers/data_request_mailer.rb
```diff
@@ -5,11 +5,12 @@ class DataRequestMailer < ApplicationMailer
   helper :application
 
   def data_request_email
+    I18n.locale = params[:locale]
```
emmabckstrm, line 8: Not sure this is the best way to go but it didn't seem to do it automatically? 🤔

PR #494, opened 2021-04-27
Cookies policy added for compliance

kallenilver:
I added this to check all the compliance boxes I think we need


emmabckstrm (approved):
Great! 👏

PR #496, opened 2021-04-30
New reported data updates existing db row when survey

emmabckstrm:


emmabckstrm:
[When survey mode the reported data is updated, instead of creating many new rows](https://trello.com/c/Z5zipNcl/617-when-survey-mode-the-reported-data-is-updated-instead-of-creating-many-new-rows)

PR #500, opened 2021-05-04
Calculator field type - using units as answer alternatives

emmabckstrm:


emmabckstrm:
[Possibility to create questions with multiple choice / radio buttons?](https://trello.com/c/c1v8fIcV/606-possibility-to-create-questions-with-multiple-choice-radio-buttons)

PR #503, opened 2021-05-11
Export reported datas as csv from admin

emmabckstrm:


emmabckstrm:
[Export reported data .csv](https://trello.com/c/POUtA9E7/610-export-reported-data-csv)

PR #504, opened 2021-05-13
Radio buttons in business product user form

emmabckstrm:


emmabckstrm:
[Radio buttons in user form](https://trello.com/c/b1qYnedJ/624-radio-buttons-in-user-form)

app/views/reported_datas/new.html.erb
```diff
@@ -19,9 +20,11 @@
...
           <% reported_datas.each do |reported_data| %>
             <%= fields_for 'reported_datas[]', reported_data do |reported_data_fields| %>
-              <%= reported_data_fields.hidden_field :calculator_field_id %>
-              <%= reported_data_fields.hidden_field :data_request_id %>
-              <%= (reported_data_fields.hidden_field :id) if @data_request.survey %>
+              <input type="hidden" name="<%= "reported_datas[#{field_index}][calculator_field_id]" %>" value="<%= reported_data.calculator_field_id %>">
+              <input type="hidden" name="<%= "reported_datas[#{field_index}][data_request_id]" %>" value="<%= reported_data.data_request_id %>">
+              <% if @data_request.survey %>
+                <input type="hidden" name="<%= "reported_datas[#{field_index}][id]" %>" value="<%= reported_data.id %>">
+              <% end %>
```
emmabckstrm, line 27: If this is possible to do with rails, I'm all ears. But this is required in order to have sets of radio buttons across the different question fields - having them have the same name `reported_datas[][unit]` makes all the radio buttons belong to the same set. Another advantage of this is that it uniforms the params between survey and non survey data requests (both are hashes instead of hash and array respectively)

PR #506, opened 2021-05-17
Submit empty answers

emmabckstrm:


emmabckstrm:
[What about submitting the data collection form without answering all questions?](https://trello.com/c/7m6wOZ3z/613-what-about-submitting-the-data-collection-form-without-answering-all-questions)

PR #510, opened 2021-05-31
Survey boolean on calculator

emmabckstrm:


emmabckstrm:
[Possibility to add multiple answers to the same question "add all the different server types and their server"](https://trello.com/c/C7DNrvYu/615-possibility-to-add-multiple-answers-to-the-same-question-add-all-the-different-server-types-and-their-server)

PR #517, opened 2021-06-11
Introduce new price for new subscribers

stefanlindbohm:
Step one will be introducing the new plans and using those for new subscriptions. WIP.

## Todo (this PR)

- [x] Fix tests.
- [x] Confirm prices (and actually change USD & CAD)
- [x] Review consequences of these changes for any undefined or unwanted behavior.

## Upcoming steps (later PR's)

- Add views for current subscribers to choose what plan they want to continue on (#518)
- Send email informing about price changes
- Add & run batch job to migrate old plans to new, taking choices made into account.
- Cleanup old plans & related code.


app/models/subscriptions/subscription_month.rb
```diff
@@ -0,0 +1,32 @@
...
+    validates_presence_of :start_at, :co2e
+    validates :payment_type, inclusion: { in: %w[CardCharge Subscriptions::ReferralCode] }
+
+    def self.create_from_stripe_invoice_line!(stripe_invoice_line, charge)
+      create!(
+        price: stripe_invoice_line.amount,
+        currency: stripe_invoice_line.currency,
+        start_at: Time.at(stripe_invoice_line.period.start),
+        co2e:
+          if (co2e = stripe_invoice_line.price.metadata['co2e']).present?
```
stefanlindbohm, line 20: switch to `if product = 2021`

app/models/greenhouse_gases.rb
```diff
@@ -4,6 +4,16 @@ class GreenhouseGases
...
 
   CONSUMER_PRICE_PER_TONNE = {
+    Currency::AUD => Money.new(9_50, :aud),
+    Currency::CAD => Money.new(6_15, :cad),
+    Currency::DKK => Money.new(44_00, :dkk),
+    Currency::EUR => Money.new(6_00, :eur),
+    Currency::GBP => Money.new(5_00, :gbp),
+    Currency::SEK => Money.new(60_00, :sek),
+    Currency::USD => Money.new(4_70, :usd)
+  }.freeze
```
stefanlindbohm, line 14: verify before deploy

app/models/stripe_extensions/plan.rb
```diff
@@ -2,40 +2,6 @@
...
-      end
-
-      def product_name(monthly_amount)
-        "Climate Offset #{monthly_amount.amount} #{monthly_amount.currency.iso_code} Monthly"
-      end
-    end
-
     def monthly_amount
       Money.new(amount, Currency.from_iso_code(currency))
     end
```
stefanlindbohm, line 7: is this used now? maybe remove

app/models/subscriptions/subscription_month.rb
stefanlindbohm: Actually, I'll just add a comment to clarify this. So we don't need to change the code next time we change prices :).

app/models/stripe_extensions/plan.rb
stefanlindbohm: Removed this

app/models/greenhouse_gases.rb
stefanlindbohm: now updated with final numbers

stefanlindbohm:
@emmabckstrm @kallenilver Feeling as this should be done now. I've created the `offsetting_subscription_2021` in Stripe production, so should be good to deploy! Merge at will :).

PR #521, opened 2021-06-21
Dont require checkbox to signup

emmabckstrm:


emmabckstrm:
[Update sign up page: Remove check box to an info text below sign up button and not have button be disabled](https://trello.com/c/UGhxcO1i/637-update-sign-up-page-remove-check-box-to-an-info-text-below-sign-up-button-and-not-have-button-be-disabled)

PR #526, opened 2021-07-02
Make reset password token expired error more human

emmabckstrm:


emmabckstrm:
[Make the error message "Reset password token is invalid" more clear to the user](https://trello.com/c/22jD5bV2/604-make-the-error-message-reset-password-token-is-invalid-more-clear-to-the-user)

PR #536, opened 2021-07-14
Using TailwindCSS in admin

emmabckstrm:
Updating the styling of admin view to use TailwindCSS instead of legacy css Bootstrap. All views are updated.

As we no longer have separate CSS `admin` vs `main`, I moved the CSS files to `shared` folder


emmabckstrm:
[Add TailwindCSS to admin](https://trello.com/c/7aielkkl/651-add-tailwindcss-to-admin)

PR #537, opened 2021-07-14
Handle user when stripe customer is deleted

emmabckstrm:
To protect ourselves from weird bugs from old users I have added checks for whether a user's stripe customer has been deleted.


stefanlindbohm:
Very nice! ✨ :shipit:

PR #541, opened 2021-08-09
Catch-all review from things I missed

stefanlindbohm:
Swooping in way late with reviews on a few things from the spring. Posting this here with simple notes and no working changes just to see what you agree with & not. I'm happy to work on the actual changes in separate PR's on things we agree on!


app/controllers/admin/dashboard_controller.rb
```diff
@@ -13,6 +13,19 @@ def index
...
+    #    isolated thing. (But of course ideally not something we should have to
+    #    do. A background job would be reasonable.)
+    #  - The logic itself is what I would consider domain logic and should live
+    #    in app/models.
+    #  - Something about refreshing a token and not getting an updated one felt
+    #    weird to me, and looking at the documentation quickly too, I'm pretty
+    #    sure that for the refresh to serve any purpose, we would need to switch
+    #    to using the token returned by the refresh_access_token call. The point
+    #    should be to rotate access tokens in case someone gets their hands on
+    #    one.
```
emmabckstrm, line 28: I agree on your first two points. I threw in this button very quickly as it needed to be done and we didn't have background jobs. I think I didn't put it in its own controller/model was because I felt it was temporary.

Regarding the documentation it says "Refreshing a long-lived token makes it valid for 60 days *again*." from https://developers.facebook.com/docs/instagram-basic-display-api/guides/long-lived-access-tokens 

doesn't that sound as though it will be the same token? However when doing the refresh curl I do get another one back. I guess we'll see on Aug 21, when the original would go out (ish I think I put a couple of days beforehand to have some slack in case the date would be on a weekend)

emmabckstrm (commented):
I put some answers into comments, ugh, I get stressed just by looking at this code -> meaning I was probably more stressed than I thought before vacay

app/controllers/lifestyle_footprints_controller.rb
```diff
@@ -42,6 +42,9 @@ def destroy
     redirect_to dashboard_path
   end
 
+  # REVIEW_NOTE: This doesn't seem to me that it needs to be a custom method.
+  # Could just be standard `update`, no? (And params would be
+  # `footprint_update_params` or similar.
```
emmabckstrm, line 47: makes sense. i guess i had some reasons for doing it like this, i think it was something about making it clear that you only update the name and nothing other, but 🤷‍♀️

PR #542, opened 2021-08-12
Remove instagram feed from startpage

emmabckstrm:
Due to the this feature not properly working (since the release before vacation) I'm removing.


stefanlindbohm:
🧹 👌

PR #544, opened 2021-08-18
User search view in admin handles when a user doesn't have a lifestyle footprint

emmabckstrm:
Prevents old users from creating trouble.


emmabckstrm:
[User search crashes](https://trello.com/c/LUXnXC7p/652-user-search-crashes)

PR #547, opened 2021-08-25
Admin content menu handles multiple actions

emmabckstrm:


emmabckstrm:
[handle many actions in admin header](https://trello.com/c/5AIzF3Ka/668-handle-many-actions-in-admin-header)

PR #549, opened 2021-08-27
Admin user actions

emmabckstrm:


app/controllers/admin/users_controller.rb
```diff
@@ -2,40 +2,65 @@
...
-        if params[:search_query].length >= min_chars_in_search_query
+        if params[:search_query].length >= MIN_CHARS_IN_SEARCH_QUERY
           @users = User.search_email(params[:search_query], @max_entries)
         else
-          redirect_to admin_users_path, notice: "Please search with at least #{min_chars_in_search_query} characters"
+          redirect_to admin_users_path, notice: "Please search with at least #{MIN_CHARS_IN_SEARCH_QUERY} characters"
         end
       else
         @users = []
       end
```
stefanlindbohm, line 19: I'd like to suggest a refactoring of this to avoid the pyramids:

```ruby
      if params[:search_query].present? && params[:search_query].length < MIN_CHARS_IN_SEARCH_QUERY
        redirect_to admin_users_path, notice: "Please search with at least #{MIN_CHARS_IN_SEARCH_QUERY} characters"
        return
      end

      @users = users_from_search(params[:search_query], @max_entries)

...

    def users_from_search(query, max_entries)
      return [] if query.blank?
      
      User.search_email(query, max_entries)
    end
```

app/controllers/admin/users_controller.rb
```diff
@@ -2,40 +2,65 @@
...
     end
 
+    def cancel_subscription
+      subscription_manager = Subscriptions::StripeSubscriptionManager.new(@user)
+
+      notice = if @user.active_subscription? && subscription_manager.cancel
+                 'Subscription successfully cancelled'
+               else
+                 'Oups! Something went wrong!'
```
stefanlindbohm, line 40: What went wrong is available on `subscription_manager.errors`. Would be nice to include that in the message

app/controllers/admin/users_controller.rb
```diff
@@ -2,40 +2,65 @@
...
+               end
+
+      redirect_to admin_user_path(@user.id), notice: notice
+    end
+
+    def destroy
+      notice = if @user.deactivate
```
stefanlindbohm, line 47: We would need tests to verify that this actually returns false in cases where things go wrong. It seems right now that the only path that this would return false is when updating of the attributes on `User` somehow fails (but we're not verifying that the cancellation of the subscription actually happens for example). Another option is to make the method always crash if errors are encountered and rename it to `deactivate!`. Either way should have tests as this could otherwise lead to some hard to debug failure modes.

app/models/user.rb
```diff
@@ -137,6 +143,30 @@ def three_months_since_last_card_charge?
     last_card_charge.created_at < (Date.today - 3.months)
   end
 
+  def deactivate
```
stefanlindbohm, line 146: I think we should verify here that a second run of this method doesn't actually do anything. Right now running this twice on the same user would create a recursive email, right? :)

stefanlindbohm (commented):
I like the general solution! Some changes in details and tests to verify that the right things happen would be good though :).

app/views/admin/users/index.html.erb
```diff
@@ -1,52 +1,54 @@
-<%= render 'admin/shared/notice_banner' %>
+<div class="space-y-3">
```
emmabckstrm, line 1: This whole file looks crazy, but I basically just added the parent `<div>` here to have proper spacing

emmabckstrm:
@stefanlindbohm great review! I think I made it work now - at least I learnt a lot about error handling! :D

app/models/user.rb
```diff
@@ -138,9 +146,55 @@ def three_months_since_last_card_charge?
...
 
   def subscription_end_at_from_stripe(stripe_subscription)
     Time.at(stripe_subscription.ended_at) if stripe_subscription.ended_at.present?
   end
+
+  def handle_errors_and_return_status
+    yield
+
+    true
+  rescue Stripe::CardError, Stripe::InvalidRequestError => e
```
stefanlindbohm, line 191: `CardError` shouldn't happen here so we shouldn't catch it. It could happen within `SubscriptionManager.cancel` but that's already handled in there :).

stefanlindbohm (approved):
Very very nice work 👏

PR #550, opened 2021-09-01
Email validation when signing up

emmabckstrm:
I have tested the regex as follows:
Screenshot 2021-09-01 at 14 11 37

The reason for updating the formatting of the auto error messages from Activerecord is that it by default show the field that's causing the error (see image below). I removed the "Email" prefix

Screenshot 2021-09-01 at 12 43 26


emmabckstrm:
[Email validation when signing up](https://trello.com/c/FrAPumnk/684-email-validation-when-signing-up)

nanditagoclimate (approved):
Looks good. 🎉

PR #552, opened 2021-09-03
Connect flight offset to user

nanditagoclimate:
**PR Description:**
1. When user offsets their flight, we tag user_id along with their email in Flight offset table
2. Add a link on dropdown (if user has offset flights) when the user can view all the flights they've offset.
3. Rake task to backfill user_ids for already created flight offsets
4. Auto populate email in Offset flight emissions if user logged in 

**Post Deployment steps:**
1. Database  Migration to add user_id column to flight offset
`rails db:migrate`
2.  Data migration to backfill user id with email match in flight offset
`rake update_user_id_in_flight_offset`


stefanlindbohm (commented):
Nicely put together! I'm a structure person so of course I'll have some comments on how to structure the code 🙈

app/controllers/dashboard_controller.rb
```diff
@@ -22,6 +22,12 @@ def show
     @projects = Project.all.order(created_at: :desc).limit(5)
 
     @total_number_of_footprints = LifestyleFootprint.where(user_id: current_user.id).count
+
+    @user_has_flight_offsets = current_user.flight_offsets?
+  end
+
+  def my_flight_offsets
+    @my_flight_offsets = FlightOffset.where(user_id: current_user.flight_offsets)
   end
```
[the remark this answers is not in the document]
emmabckstrm: I was thinking something similar! You can see how it's done for viewing all your previous footprints at `/calculator/results` (requires you to have multiple footprints thought, I think)

app/views/dashboard/my_flight_offsets.html.erb
```diff
@@ -0,0 +1,34 @@
...
+        <thead>
+          <tr>
+            <th>Date:</th>
+            <th>Price:</th>
+            <th>Tonnes of CO2e offsetted:</th>
+          </tr>
+        </thead>
+
+        <tbody>
+          <%@my_flight_offsets.each do |flight_offset| %>
```
emmabckstrm, line 20: Super picky, but I'm in favor of not prefixing user related code stuff with `my_`, even though I know there is some code in our codebase that does this :)

app/views/dashboard/show.html.erb
```diff
@@ -97,6 +97,11 @@
           class: 'link whitespace-nowrap'
         %>
       </div>
+      <% if @user_has_flight_offsets %>
```
emmabckstrm, line 100: Could be checked with `current_user.flight_offsets?`

app/views/dashboard/show.html.erb
```diff
@@ -97,6 +97,11 @@
...
+        <div class="inline-block">
+          <a class="link" href="/dashboard/my_flight_offsets"><%= t 'dashboard.flight_offsets' %></a>
+        </div>
```
emmabckstrm, line 103: I know the navigation in the dashboard is not perfect at all, and needs rework - however, I do feel that this might not be the best place for this link. It looks like in your screenshot that you have not multiple lifestyle footprints on this user, so you're not seeing the "See previous footprints" link. It's already a crammed place, and in my opinion I would keep the links below the footprint related to the footprint or the subscription.

My suggestion is to move this link up to the drop down menu where we already have a link to "Flight offsets" and put "My flight offsets" above/below to it

db/schema.rb
```diff
@@ -197,6 +197,7 @@
     t.text "currency"
     t.text "payment_intent_id"
     t.datetime "paid_at"
+    t.bigint "user_id"
```
emmabckstrm, line 200: Shouldn't this have a foreign key constraint?

emmabckstrm (commented):
Great work Nandita! 🎉  I have some comments as well, mainly on some details

app/controllers/dashboard_controller.rb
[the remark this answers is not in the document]
nanditagoclimate: Having a `UserFlightOffsetsController` makes sense. 
My initial thoughts were that since we show this on dashboard, we could add user specific data end points there but in case we want to move it to a different view later on, it'll make it so much easier with a dedicated controller. 👍

app/controllers/flight_offsets_controller.rb
```diff
@@ -21,6 +21,9 @@ def create
       params.require(:flight_offset).permit(:co2e, :currency, :email)
     )
 
+    user = current_user || User.find_by_email(params['flight_offset']['email'])
+    @offset.user_id = user.id if user
```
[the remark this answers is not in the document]
nanditagoclimate: I am not a fan of callbacks unless then are absolutely necessary. Testing callbacks may become complicated. Moreover, this particular case doesn't fit in under a callback I feel. It is not tied to a db action. The only reason to put this in a callback would be to cover it under tests. So, I prefer to leave it here itself. Thoughts please?

app/views/dashboard/my_flight_offsets.html.erb
nanditagoclimate: I shall rename this to user_flight_offsets as Stefan's suggested. I couldn't come up with a better name then.

db/schema.rb
nanditagoclimate: As a good practice, in cases like ours, it's better to not have foreign key constraints on database. Especially in Rails, if we can set up relationships as associations. The reason being, in case we decide to migrate to a different database in future, not having this foreign key dependency will make it simpler. Please correct me if i'm wrong.

app/views/dashboard/show.html.erb
nanditagoclimate: Yes, can be. 
But personally I prefer not having explicit checks like this on erb later and instead pass them from controller.

app/views/dashboard/show.html.erb
nanditagoclimate: You're right. I don't see "See previous footprints" link on my local. I shall move this to the drop down! Thanks.

app/controllers/flight_offsets_controller.rb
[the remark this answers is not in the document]
stefanlindbohm: I'm not entirely convinced that callbacks need to be absolutely necessary, but I know they can become a bit complicated so I'm not going to push for it. Maybe another solution is to move this logic into a `User#infer_user(User)` method? Both makes it more explicit and allows testing, but without callbacks :)

db/schema.rb
stefanlindbohm: I agree with @nanditagoclimate that it's enough to let ActiveRecord keep track of associations. But I would opt for using the `references` helper in the migration to let Rails setup things using current defaults.

app/controllers/flight_offsets_controller.rb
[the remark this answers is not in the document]
nanditagoclimate: I am not too sure I understand this. Do you mean we add an instance method on user that simply returns user id if user is present? Or actually updates the user_id in flight_offset object post its creation if user is present?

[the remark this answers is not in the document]
stefanlindbohm: The latter :). I'm thinking move this to a method in `User` like this:

```ruby
def infer_user(current_user)
  return if user_id.present?

  user_id = (current_user || User.find_by_email(email))&.id
end
```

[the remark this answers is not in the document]
nanditagoclimate: I have added something similar. Removed the "current_user ||.." check. 
Reason being, if a logged in user enters a different email, we will be tagging the user_id incorrectly. 
Please check and let me know your thoughts.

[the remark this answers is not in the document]
stefanlindbohm: That makes sense! Sorry for being unclear in my feedback–I really didn't mean this as a big thing or a showstopper!

app/models/flight_offset.rb
```diff
@@ -90,6 +90,10 @@ def price=(price)
     super(price)
   end
 
+  def self.infer_user_id(email)
+    User.find_by_email(email)&.id
+  end
+
```
stefanlindbohm, line 96: Just one minor details here. `email` is already an attribute here, so we don't need to take it as a parameter to this method. Right?

nanditagoclimate: Since this is a class method and is invoked before the flight offset object is created, we will need to explicitly pass the email as a param.
If we invoked this method on the flight offset object itself, then we wouldn't have to pass it as a param.

nanditagoclimate: Made the change. Makes more sense and looks cleaner. 👍 thanks for pointing it out!

stefanlindbohm: Yes, that's what I meant! I somehow missed the `self.`, sorry again for the back and forth!

PR #553, opened 2021-09-03
Change user email

emmabckstrm:
Screenshot 2021-09-07 at 15 00 05
Screenshot 2021-09-07 at 15 00 12


app/controllers/admin/users_controller.rb
```diff
@@ -20,6 +20,21 @@ def show
...
 
+    def edit
+    end
+
+    def update
+      if @user.update(user_params)
+        Stripe::Customer.update(
+          @user.stripe_customer.id,
+          email: @user.email
+        )
```
stefanlindbohm, line 31: I'll sneak in a comment in there that this should happen in `User` because the user themself can change their email and it would be great if Stripe was updated in those cases too :). I'll recommend either an `around_save` that checks if the `email` attribute is being changed and if so calls Stripe after the update is committed, or simply an `after_save` that always calls Stripe (the latter would be simpler but less performant of course).

emmabckstrm:
[Change email for users also change in stripe](https://trello.com/c/hLA1DYTP/704-change-email-for-users-also-change-in-stripe)

app/controllers/admin/users_controller.rb
stefanlindbohm: @emmabckstrm Oh yeah, smart there :D

app/controllers/admin/users_controller.rb
```diff
@@ -20,6 +20,17 @@ def show
...
 
+    def edit
+    end
+
+    def update
+      if @user.update(user_params)
+        redirect_to [:admin, @user], notice: 'User was successfully updated.'
+      else
+        render :edit, notice: "Oups, something went wrong! #{@user.errors.full_messages}"
+      end
```
stefanlindbohm, line 31: So I'll be the parrot. I'd prefer to do this as a guard clause like: `unless @user.update(user_params); ... return`. Saves the `else` clause and clarifies at a glance which is the "normal" behavior.

app/models/user.rb
```diff
@@ -180,6 +182,22 @@ def deactivate
 
   private
 
+  def update_email_in_stripe
+    yield
```
stefanlindbohm, line 186: Since you're not doing anything before yielding, you can switch this to an `after_update` and remove this line.

app/models/user.rb
```diff
@@ -180,6 +182,22 @@ def deactivate
...
+    return unless errors.blank? && saved_change_to_attribute?('email')
+
+    begin
+      Stripe::Customer.update(
+        stripe_customer.id,
+        email: email
+      )
+    rescue Stripe::InvalidRequestError => e
+      logger.error(e)
+      Raven.capture_exception(e)
```
stefanlindbohm, line 197: 👌

stefanlindbohm (commented):
Nicely done! 👏

emmabckstrm:
@stefanlindbohm I've agreed with your changes. Do you want to look at it again - or ok to deploy?

app/controllers/admin/users_controller.rb
```diff
@@ -20,6 +20,18 @@ def show
       @total_number_of_footprints = LifestyleFootprint.where(user_id: @user.id).count
     end
 
+    def edit
+    end
+
+    def update
+      unless @user.update(user_params)
+        render :edit, notice: "Oups, something went wrong! #{@user.errors.full_messages}"
```
nanditagoclimate, line 28: Oops* ?

nanditagoclimate (approved):

app/controllers/admin/users_controller.rb
stefanlindbohm: I have a pet peeve about this, too. If something goes wrong we should apologize to the user and/or help them fix the problem, not say ”oops” :)

stefanlindbohm:
> @stefanlindbohm I've agreed with your changes. Do you want to look at it again - or ok to deploy?

Just go! :shipit:

app/controllers/admin/users_controller.rb
emmabckstrm: @nanditagoclimate @stefanlindbohm Lucky it's only we that would have seen this and not our users ;)

emmabckstrm: Even though I don't think this type of wording (as long as it's spelled correctly) is too bad. But I'm not interested into putting energy into discussing it :)

PR #555, opened 2021-09-07
Cleanup ghg formatting

stefanlindbohm:
Changes:

- Doesn't switch automatically between kgs and tonnes. Do that explicitly instead.
- Now uses delimiters and separators from current locale.

Specifically to adress the request to make admin dashboard more legible:

Screen Shot 2021-09-07 at 15 19 57

### DIFF
--- a/Guardfile
+++ b/Guardfile
@@ -61,7 +61,7 @@ group :javascript do
   end
 end
 
-guard :brakeman, run_on_start: false do
+guard :brakeman, quiet: true do
   watch(%r{^app/.+\.(erb|haml|rhtml|rb)$})
   watch(%r{^config/.+\.rb$})
   watch('Gemfile')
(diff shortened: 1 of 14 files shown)


Guardfile
```diff
@@ -61,7 +61,7 @@ group :javascript do
   end
 end
 
-guard :brakeman, run_on_start: false do
+guard :brakeman, quiet: true do
```
emmabckstrm, line 64: What does this do?

emmabckstrm (approved):
Nice! Just had a question regarding your change to brakeman

Guardfile
stefanlindbohm: It prints from Brakeman less each time Guard triggers a re-run from a saved file. Docs: https://github.com/guard/guard-brakeman#list-of-available-options

PR #556, opened 2021-09-07
Tweak cancellation buttons

stefanlindbohm:
I think the current situation is a little bit confusing. This change aims to:

- Draw less attention to cancelling your subscription.
- Clarify that the danger is actually happening when clicking cancel in
  the modal.

## Before
Screen Shot 2021-09-07 at 15 38 40
Screen Shot 2021-09-07 at 15 39 00

## After
Screen Shot 2021-09-07 at 15 38 40
Screen Shot 2021-09-07 at 15 38 44


emmabckstrm (approved):
Nice!
I think the first screenshot is wrong thought- it's not showing a difference

stefanlindbohm:
> Nice!
> I think the first screenshot is wrong thought- it's not showing a difference

Oops, yeah I added the wrong screenshot 🙈

PR #558, opened 2021-09-08
Keep track of subtotals

stefanlindbohm:
This implements a few changes with the goal of better availability of various subtotals/taxes for our products.

- Adds `price_incl_taxes` to replace `price` for `SubscriptionMonth`, `GiftCard` and `FlightOffset`. `price` is now the price without taxes. Also adds `vat_amount` which is the VAT portion of the total price.
- Adds subtotals for offsetting, consulting & product portions of total `Invoice` amounts. Keeping this is a manual task like the total amount is today. Adding this data will be a manual task, too, that I'm planning to do once this is merged.
- Changes the meaning of `cost_in_sek` for `Project` to mean cost without any VAT. This is actually a necessity because VAT rates differ depending on which country our supplier invoices in (VAT paid, reverse VAT responsibility, etc). Updating this data will be another manual task that I'm also planning to do once this is merged.
- Refactors a few things, most notably creates a `HasMoneyAttributes` with shared logic to turn attribute accessors to working straight with `Money` objects, as well as changing all touched money attributes to use this translation.

## Deployment

This needs to be deployed in three steps:

1. Migrations & some refactoring in preparation (#567)
2. Backwards-compatible use of new columns (#568)
3. Remove backwards compatibility & use new columns fully (this PR)

## Review

All commits are included in this PR, so a full review can be done here. Step 1 & 2 PR's include the same commits, but let's make sure those in between steps are good to go live too :).


app/models/concerns/has_money_attributes.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+##
+# Adds the +money_attribute+ class method to a model class. This helper creates
+# accessor methods for an attribute of money type, translating between `Money`
```
stefanlindbohm, line 5: Fix formatting

db/migrate/20210913134601_migrate_invoices_projects_money_data.rb
```diff
@@ -0,0 +1,21 @@
+class MigrateInvoicesProjectsMoneyData < ActiveRecord::Migration[6.0]
```
stefanlindbohm, line 1: Disable transaction

app/models/concerns/has_money_attributes.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+##
+# Adds the +money_attribute+ class method to a model class. This helper creates
+# accessor methods for an attribute of money type, translating between +Money+
+# objects and +ActiveRecord+ types.
+#
+# Defined attributes are added to an anonymous module to allow the class to
+# overwrite methods with further functionality.
+module HasMoneyAttributes
```
kallenilver, line 10: This is already added in step1? I understood from the description that all three steps should be deployed, but now I see that this PR contains the other steps?

app/models/flight_offset.rb
```diff
@@ -5,21 +5,27 @@
 class FlightOffset < ApplicationRecord
   class InvalidPaymentIntentError < StandardError; end
 
+  include HasMoneyAttributes
+
   attribute :currency, :currency
   attribute :co2e, :greenhouse_gases
+  money_attribute :price, :currency
+  money_attribute :vat_amount, :currency
+  money_attribute :price_incl_taxes, :currency
```
kallenilver, line 14: Is taxes something other than VAT?

kallenilver (approved):
Looks good, but I don't have the possibility to deep dive into this right now :)

emmabckstrm (approved):

app/models/flight_offset.rb
stefanlindbohm: Not right now, no. But I felt this was a more natural term to mean "total of everything". It could be other taxes in the future, too. But primarily just because VAT is an implementation detail in relation to "get me the full price to pay" :).

app/models/concerns/has_money_attributes.rb
stefanlindbohm: Yeah, this PR includes all commits. So merging this after steps 1 & 2 will only add the last commits. I wanted one PR for the full review. Maybe something to clarify or do otherwise in similar PR's in the future.

app/models/money.rb
```diff
@@ -23,10 +25,22 @@ def <=>(other)
...
+  end
+
   def *(other)
     Money.new((subunit_amount * other).round, currency)
   end
 
+  def /(other)
+    Money.new((subunit_amount / other).round, currency)
+  end
+
```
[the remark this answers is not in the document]
stefanlindbohm: Because multiplication and division work with numbers, while addition and subtraction work with other `Money` objects. So far we're only checking matching currencies when two `Money`s are involved. Adding type checks would be a nice addition though, but it's already quite apparent what's wrong when you do that part wrong (it will crash for other reasons :) ).

PR #563, opened 2021-09-14
Ensure new category, field added gets added at the end instead

nanditagoclimate:
When a new field is added to a category, it takes the first position. Ideally, it should get added at the end.
Same happens when a new category is added.

This PR ensures right behaviour.


nanditagoclimate:
The solution is too hacky.
I would like to redo certain parts of the code:
1. Change the way ordering is handled. 
2. Some changes on update logic as well.

app/models/business_calculators/calculator_field.rb
```diff
@@ -10,6 +10,14 @@ class CalculatorField < ApplicationRecord
     validate :units_exists
     validate :correct_field_type
 
+    scope :order_as, lambda { |ids|
```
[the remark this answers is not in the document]
nanditagoclimate: Need help. SOS

nanditagoclimate:
OMG. So many CI/CD bugs. I have resolved most of them. But i'll make sure my future PRs are clean.

app/models/business_calculators/calculator_category.rb
```diff
@@ -16,8 +16,20 @@ class CalculatorCategory < ApplicationRecord
         attributes['label'].blank?
       }
 
+    scope :order_as, lambda { |ids|
```
nanditagoclimate, line 19: This logic is duplicated in both field and category model. I couldn't do better. SOS

emmabckstrm: I don't 100% remember what you and @stefanlindbohm talked about on today's standup: was it about this logic here and it's not possible to use a helper here? if that is some Rails magic that prevents it then I have no idea on how to go around it :O

stefanlindbohm (approved):
I think this is a clear improvement and good enough for now. Great job!

Two ideas on how this could be done in a way with less duplication:

- We already have logic in the front end to change the sorted comma separated list of ID's. Maybe one way would be to add this new ordering behavior to that code?
- If we are to keep sorting logic in the back end code, one way to avoid duplication would be to design it as a concern. Something like `HasOrderedAttributes` that only needs to adapt to which attribute is the order field and which is the association that is being sorted.

Not sure if we need to act on this now. I'll leave that up to you if you feel inspired to take another stab or just want to merge it :). This PR as is is good too 👍 .

app/helpers/business_calculators_helper.rb
```diff
@@ -0,0 +1,47 @@
...
+      @calculator.update(category_order: @calculator.categories.pluck(:id))
+    # New category added or a category is deleted
+    elsif @calculator.category_order.sort != @calculator.categories.pluck(:id).sort
+      new_categories = @calculator.categories.pluck(:id) - @calculator.category_order
+      deleted_categories = @calculator.category_order - @calculator.categories.pluck(:id)
+      categories = @calculator.category_order + new_categories - deleted_categories
+      @calculator.update(category_order: categories)
+    # When categories are reordered
+    else
+      @calculator.update(category_order: @category_order)
```
stefanlindbohm, line 26: Is this changing anything? Looks to me like it's setting the same value back.

app/controllers/admin/business_calculators_controller.rb
```diff
@@ -26,21 +27,28 @@ def edit
...
-      else
-        render :new
+      begin
+        @calculator = BusinessCalculators::Calculator.create!(calculator_params.except('categories_attributes'))
+        calculator_params['categories_attributes'].each_value do |new_category|
+          next unless new_category['name'].present?
+
+          category = @calculator.categories.create!(name: new_category['name'])
+          new_category['fields_attributes'].each_value do |new_field|
+            category.fields.create!(new_field) if new_field['label'].present?
```
emmabckstrm, line 37: Maybe stupid question, but, shouldn't we specify category id here? Or does it do that automatically because of nested fields? 🤔

emmabckstrm:
Oh, I forgot to ask this as well - are you planning on cleaning up the commits messages? 

And last thing: if you want to ignore CodeClimate you click on `Details` and log in (I'm using Github login) and you should have a button that says "Approve" and it should stop complaining

app/models/business_calculators/calculator.rb
```diff
@@ -16,6 +17,10 @@ class Calculator < ApplicationRecord
 
     after_create :set_default_status
 
+    def ordered_categories
+      BusinessCalculators::CalculatorCategory.where(calculator_id: id).order_as(category_order)
+    end
+
```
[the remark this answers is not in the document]
nanditagoclimate: .categories will continue to function as before without the ordering part and this happens with association defined on the model.
If we rename ordered_categories to just categories then we will be over riding the association + nested attributes logic and that in turn will require us to change a lot of code. Including creation and update logic.

You're right. I will need to update this in .categories in  report_area.rb and reported_datas_controller.rb and other places.
I couldn't come up with a default way of ordering this without the sql injection part that will be deprecated soon.

app/controllers/admin/business_calculators_controller.rb
nanditagoclimate: Yes! It happens through association.

nanditagoclimate (commented):
I think the commits can continue to be as is. I will anyway be `squash merging` this into master once ready so it'll be a single commit there. Logs won't look bad.

nanditagoclimate:
I tried creating a concern and included it in the modules but within the scope or even directly I am unable to invoke it from the view. When I run the queries in the console directly it works fine, so did the helper. But through the app, when we invoke ordered_categories or ordered_fields like what we do above, the method is undefined. 

I shall leave the PR as is and have `approved` the duplicate code warning on CC. :)

app/models/business_calculators/calculator.rb
```diff
@@ -58,18 +58,26 @@ def duplicate # rubocop:disable Layout/MethodLength
...
           category.fields.each do |field|
             new_field = field.dup
             new_field.category_id = new_category.id
             new_field.save
+            field_order << new_field.id
           end
+          new_category.update!(field_order: field_order)
         end
+        new_calculator.update!(category_order: category_order)
```
emmabckstrm, line 80: 😍 woop woop! Great solution!

emmabckstrm (approved):

PR #564, opened 2021-09-16
Reported datas excel export

emmabckstrm:
This PR is easiest to look at commit by commit.

- b99c0eb: Simple but very important bugfix
- cdd2942: The meat of the PR. Update the admin view and add excel export (see screenshots).
- 1e52dac: Accidentally let rubocop auto fix some styling issues :) 

Screenshot 2021-09-16 at 11 41 36

Screenshot 2021-09-16 at 11 40 38


Gemfile.lock
```diff
@@ -413,4 +420,4 @@ RUBY VERSION
    ruby 2.6.6p146
 
 BUNDLED WITH
-   2.1.4
+   2.2.27
```
emmabckstrm, line 423: Does this look like it should? I've never added a gem before :D

app/controllers/admin/reported_datas_controller.rb
```diff
@@ -2,56 +2,54 @@
...
-      @reported_datas = format_data(reported_datas)
     end
 
-    def show
-      @calculator_field = BusinessCalculators::CalculatorField.find(@reported_data.calculator_field_id)
-      data_request = DataRequest.find(@reported_data.data_request_id)
-      @report = ClimateReports::ReportArea.find(data_request.report_area_id).report
-      @data_reporter = DataReporter.find(data_request.recipient_id)
-    end
+    def export_to_xlsx # rubocop:disable Layout/MethodLength - due to contents of array added to row, too long to have as one row
```
emmabckstrm, line 12: CodeClimate thinks that this whole method is too complicated! I have done my best to simplify it though, but couldn't do better than this

Gemfile.lock
nanditagoclimate: This should be okay. When I added a gem, bundler version was updated too.
My local bundler is on 2.2.26.

PR #566, opened 2021-09-17
Rubocop fixes

kallenilver:
Most autofixes, but a three todos added that I'm not sure if we could do something better with


emmabckstrm (approved):

stefanlindbohm:
👏

PR #571, opened 2021-09-27
Migration to rename recipient_id in data requests

nanditagoclimate:
Rename recipient_id column in Data Request table to data_reported_id cos it just sounds right and removed ambiguity. 
Post deploy:
1. Run rails db:migrate

Other PR with renamed column updates - https://github.com/goclimate/goclimate/pull/572


emmabckstrm (approved):

stefanlindbohm:
@emmabckstrm @nanditagoclimate I appreciate the effort (really do!) to do a zero-downtime deploy here, but this way of splitting migrations to a separate PR doesn't fully cover all cases. In this case, the downtime was no problem because no user is currently using this feature. For the future, here's my comments on splitting PR's to do zero-downtime deploys.

Actually, doing it like this is similar to keeping the migration in the same PR. The difference will only be whether downtime will be before or after the migration is run.

Doing it this way will have this effect:

1. Merge this PR and run deployments: Now the column name has changed but the code doesn't yet reflect that. Any code running that touches this column (or entire model actually, because of how ActiveRecord works) will now crash.
2. Merge the next PR: Here comes the code that knows about the column rename from the previous migration. This will fix the issue, but we've now run for a while with code crashing.

The key thing to keep in mind for zero-downtime deploys is that the code of any given deployment needs to be able to work fully both before and after its migrations have been run. This means that migrations that are backwards-compatible are to be deployed & run before related code is deployed (such as adding a column) and migrations that are forwards-compatible are to be run after (such as removing a column). Changes that are neither forwards- or backwards-compatible, such as renames, need to be split up into multiple steps. Usually by first creating a new column with the new name, then using both old and new columns simultaneously as a transition, then finally switching to use the new name and dropping the old column. There are many ways to do this, but here's an article mentioning a few different scenarios for more examples: https://medium.com/klaxit-techblog/zero-downtime-migrations-with-activerecord-47528abe5136.

With that said, there are cases where doing all this isn't worth it, for example if the code is only being run from our admin where a crash isn't a problem. But when it is, then we might need a few more steps than this PR :).

emmabckstrm:
@stefanlindbohm great points, thanks for such an extensive write-up! Will keep this in mind in the future :)

nanditagoclimate:
Agreed @stefanlindbohm Thank you for the explanation. Will keep this in mind going forward!

PR #573, opened 2021-09-28
Remove unit column from reported data table

nanditagoclimate:


emmabckstrm (changes requested):
This should not be removed as it's being used. I meant that the units table have at least one column that is not being used. Sorry for being unclear 🙈

nanditagoclimate:
My bad. In Units table I see `input_type` and `key`. Can I remove both of them?
From production data, I see `input_type` has only 2 distinct values.
`key` has many distinct values but I am not sure if they are useful.

nanditagoclimate:
This was an incorrect PR. Correct PR - https://github.com/goclimate/goclimate/pull/575

PR #576, opened 2021-09-30
Sidekiq, redis setup + config files

nanditagoclimate:


nanditagoclimate:
At this point and scale, it makes more sense to go with Delayed Jobs.
https://github.com/goclimate/goclimate/pull/578/

PR #578, opened 2021-10-01
Delayed jobs set up + config

nanditagoclimate:


config/initializers/delayed_job_config.rb
```diff
@@ -0,0 +1,11 @@
...
+
+Delayed::Worker.destroy_failed_jobs = false
+Delayed::Worker.sleep_delay = 60
+Delayed::Worker.max_attempts = 3
+Delayed::Worker.max_run_time = 5.minutes
+Delayed::Worker.read_ahead = 10
+Delayed::Worker.default_queue_name = 'default'
+Delayed::Worker.delay_jobs = !Rails.env.test?
+Delayed::Worker.raise_signal_exceptions = :term
+Delayed::Worker.logger = Logger.new(File.join(Rails.root, 'log', 'delayed_job.log'))
```
stefanlindbohm, line 11: In production, log files aren't read. Instead, all STDOUT will be sent to Papertrail. So let's do this config only when we're not in `Rails.env.production?`

stefanlindbohm (approved):
So nice! 🎉

PR #581, opened 2021-10-04
Draft: Move emails to background job

nanditagoclimate:
Moving all emails to background. Need to review this list and probably not make all of them asyn or we may have to set priority.


nanditagoclimate:
This PR is messed up. opening https://github.com/goclimate/goclimate/pull/583 instead.

PR #582, opened 2021-10-05
Welcome email now sent for 3D secure signup users

nanditagoclimate:
When users sign up with a 3D Secure card, welcome email not sent.
Reason:
`@manager.confirmation_required?` is true. So we only rendered confirmation_required_json.
Currently, whether the payment is successful or not, user is created and can log in to the app.

This PR moves the welcome mail delivery from create end point  to a common private method that gets invoked  after all kinds of signup: free/regular paid signup and 3D secure pay signup.


stefanlindbohm (commented):
Makes sense to simplify this by sending the email as soon as the account is anyway created, regardless of whether the subscription is successfully created or not. But then my next thought was "wait, don't we mention the subscription in the welcome email itself?" and it turns out we do.

@emmabckstrm @kallenilver What is the history on this? Why are we sending the same welcome email to both free and subscription signups?

I'd be OK with merging this as an intermediate step that's better (more consistent) than currently, but it seems we should fix this somehow too, right?

kallenilver:
> Makes sense to simplify this by sending the email as soon as the account is anyway created, regardless of whether the subscription is successfully created or not. But then my next thought was "wait, don't we mention the subscription in the welcome email itself?" and it turns out we do.
> 
> @emmabckstrm @kallenilver What is the history on this? Why are we sending the same welcome email to both free and subscription signups?

@stefanlindbohm I can't remember the history, not sure I was involved. Sounds to me now like a better alternative is to send two different emails and thereby be a little sharper in the communication. I could do the copy if you want me to?

PR #583, opened 2021-10-05
Moving admin + welcome email to delayed jobs

nanditagoclimate:
Moving admin and welcome emails to aync jobs.


stefanlindbohm (approved):
So excited for us to finally do this properly! :shipit:

PR #586, opened 2021-10-07
Make office_area mandatory and remove default value fill

nanditagoclimate:
Post deploy:
rake update_office_area_in_climate_reports


emmabckstrm (approved):
Looks good!

For your question, I assume that those that currently have office area as `nil` used the average office area option? If so, yeah, let's backfill it. I haven't really worked with this area before so this might be obvious :)

lib/tasks/update_office_area_in_climate_reports.rake
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+task update_office_area_in_climate_reports: :environment do
+  AVERAGE_OFFICE_AREA_PER_EMPLOYEE = 15 # sqm/employee
+  ClimateReport.where(office_area: nil).each do |climate_report|
```
stefanlindbohm, line 5: Doesn't matter in this case, but it's a good habit to use `each_in_batches` anytime we iterate over an unknown number of records.

PR #591, opened 2021-10-13
Add subscription cancellation email

emmabckstrm:


app/mailers/subscription_mailer.rb
```diff
@@ -28,6 +28,11 @@ def payment_failed_email
          asm: { group_id: SENDGRID_ASM_GROUP_IDS[:payment_failed] }
   end
 
+  def confirm_cancellation_email
+    mail subject: I18n.t('mailers.subscription.cancellation.subject'),
+         asm: { group_id: SENDGRID_ASM_GROUP_IDS[:really_cool_unsubscribe] }
```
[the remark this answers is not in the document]
stefanlindbohm: Yes, please fix :)

app/models/subscriptions/stripe_subscription_manager.rb
```diff
@@ -73,7 +77,7 @@ def handle_errors_and_return_status
       yield
 
       true
-    rescue Stripe::CardError => e
+    rescue Stripe::CardError, Stripe::InvalidRequestError => e
```
stefanlindbohm, line 80: We shouldn't catch invalid request errors in my opinion. This because this would only happen from bugs in our code (as opposed to externalities that could happen) and in those cases, we have no idea what went wrong and data would likely be in an unknown state anyway. So we should always get those crash reports and investigate.

app/models/user.rb
```diff
@@ -155,11 +157,13 @@ def deactivated?
 
   def deactivate
     handle_errors_and_return_status do
-      raise(StandardError::ArgumentError, 'User is already deactivated') if deactivated?
+      raise(AlreadyDeactivatedError, 'User is already deactivated') if deactivated?
```
stefanlindbohm, line 160: Nice!

app/models/user.rb
```diff
@@ -155,11 +157,13 @@ def deactivated?
...
 
       if active_subscription?
         subscription_manager = Subscriptions::StripeSubscriptionManager.new(self)
-        subscription_manager.cancel
+        unless subscription_manager.cancel
+          raise "There was an error when cancelling the subscription #{subscription_manager.errors}"
```
stefanlindbohm, line 165: Has this happened in practice? I'm a little unsure on this error handling because it seems a little speculative to me. Also, raising an error here still crashes the rendering and the user would anyway receive a HTTP 500, right? 🤔

config/locales/views/mailers/subscription.en.yml
```diff
@@ -0,0 +1,7 @@
+en:
+  mailers:
+    subscription:
+      cancellation:
+        subject: Your subscription has been cancelled
+        heading: Thank you for having contributed to stopping climate change!
```
stefanlindbohm, line 6: Maybe reword this to say "Thank you for your contributions towards stopping climate change!"? I feel using the "having" in there might be (very slightly) passive aggressive :)

config/locales/views/mailers/subscription.en.yml
```diff
@@ -0,0 +1,7 @@
...
+        text: We hereby confirm that your subscription with us has been cancelled. We are sorry to see you go, but grateful for your time with us. If you change your mind, you are always welcome back.
```
stefanlindbohm, line 7: Maybe we could also include a line saying something like "Have any questions, comments or feedback? Please get in touch with us by responding to this email."?

stefanlindbohm (commented):
Very nice to have that email!

I'm unsure about the error handling, though. Not sure it's needed at this point.

config/locales/views/mailers/subscription.en.yml
emmabckstrm: Great suggestion!

config/locales/views/mailers/subscription.en.yml
emmabckstrm: yup 👍

app/models/user.rb
emmabckstrm: I don't know! It's very speculative 😅  I'll remove it then 🙈

PR #593, opened 2021-10-15
Tweak invoice offsetting formatting

stefanlindbohm:
Now switches between kgs and tonnes based on offsetting made.


stefanlindbohm:
I'll merge this because Cissi needs it, but please review and I'll fix it with new commits. Pinged all of you, but only one review should be necessary :)

PR #594, opened 2021-10-18
Add GBP as business currency

emmabckstrm:


app/models/climate_report_invoice.rb
```diff
@@ -17,8 +17,8 @@ class ClimateReportInvoice < ApplicationRecord
 
   def calculate_from_report
     self.co2e = climate_report.calculation.total_emissions * BUFFER_FACTOR
+    self.currency = currency || Currency::SEK
```
stefanlindbohm, line 20: This line could be removed to make currency a required parameter instead. I think that would be clearer as it would force a crash if you don't include it, now that this actually has some significancy.

stefanlindbohm (approved):
Nice! Just that one small thing (see comment below).

PR #595, opened 2021-10-18
Christmas campaign page

emmabckstrm:
Restructured the files a little bit to keep the campaign pages together.

Screenshot 2021-10-20 at 10-28-37 GoClimate


app/controllers/campaigns/christmas_controller.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Campaigns
+  class ChristmasController < ApplicationController
+    def show
+      force_sv_locale
```
stefanlindbohm, line 6: This could be a `before_action` call instead. Would be a little cleaner I think.

PR #603, opened 2021-11-02
Update business price per tonne CO2

emmabckstrm:


emmabckstrm:
[Make sure business offset price is correct. It should be 75 and not 55 SEK](https://trello.com/c/Zy4Eb6Wa/747-make-sure-business-offset-price-is-correct-it-should-be-75-and-not-55-sek)

nanditagoclimate (approved):
Awesome! 🎉

PR #605, opened 2021-11-05
Handle cases where @user is nil

nanditagoclimate:


nanditagoclimate:
I'm not happy with this solution. It solves the problem at hand but this issue shouldn't have come up in the first place as SubscriptionMailer is called with user object where email is stored. But there is no email which caused this issue on database.

nanditagoclimate:
This PR will handle the bug but now fix the problem. Closing this PR. Created a bug on trello that ensures such issues won't happen in the first place. (solution TDB)

PR #606, opened 2021-11-09
Business subscriptions

emmabckstrm:
Migration in #608


emmabckstrm:
[Add new prices for business customers](https://trello.com/c/F7533QvR/754-add-new-prices-for-business-customers)

PR #614, opened 2021-11-26
Fix link in climate report emails to customers

stefanlindbohm:
Region arguments for jobs weren't properly serialized, leading to links in climate report emails having a JSON string injected into them. This was introduced in 54ef278.

Fix makes serialization of Region as jobs arguments work through the built-in serialization methods.


kallenilver (commented):
🥳

PR #619, opened 2021-11-30
PDFs with Grover

emmabckstrm:
The previous solution with WickedPDF is very memory intense and requires an expensive heroku dyno with 2,5GB RAM. This solution seems to be much more efficient and works fine on heroku with tiny dynos. Here it's implemented just for the Climate Reports PDFs as we are generating them in our admin view and manually sending them to our customers - this is a good set up for testing and making sure it works as intended.

To run on heroku, we'll need to follow the instructions [here](https://github.com/Studiosity/grover#running-on-heroku), and add buildpacks.

See more thoughts in the trello card: https://trello.com/c/8o7ZQIfG


kallenilver (approved):

nanditagoclimate (approved):
🎉

emmabckstrm:
[Discover ways of replacing current PDF solution with a more stable set up](https://trello.com/c/8o7ZQIfG/769-discover-ways-of-replacing-current-pdf-solution-with-a-more-stable-set-up)

PR #621, opened 2021-12-03
Mention email address in reset password email

emmabckstrm:


emmabckstrm:
[Reset password email mentions "recipient" instead of a name](https://trello.com/c/80wljgKP/772-reset-password-email-mentions-recipient-instead-of-a-name)

PR #623, opened 2021-12-07
Rake task for updating price of existing subscribers

emmabckstrm:
Run `rake update_price_for_existing_consumers`


nanditagoclimate:
@emmabckstrm  Please review. Since it's your PR, I can't tag you for review. 🥇

lib/tasks/update_price_for_existing_subscribers.rake
```diff
@@ -0,0 +1,36 @@
...
+  end
+end
+
+def print_logs(pricing_update_success_emails, pricing_update_failed_emails, pricing_update_declined_emails)
+  puts "Pricing updated for #{pricing_update_success_emails.count} users successfully"
+  puts pricing_update_success_emails
+  puts "Pricing update failed for #{pricing_update_failed_emails.count} users"
+  puts pricing_update_failed_emails
+  puts "Pricing update declined for #{pricing_update_declined_emails.count} users"
+  puts pricing_update_declined_emails
```
emmabckstrm, line 35: Maybe put all the counting prints at the bottom so when I run this I get an easy overview of how many users did what.

So:

```
puts "Success emails: #{pricing_update_success_emails}"
puts "Declined emails: #{pricing_update_declined_emails}"
puts "Failed emails: #{pricing_update_failed_emails}"
puts "Pricing updated for #{pricing_update_success_emails.count} users successfully"
puts "Pricing update declined for #{pricing_update_declined_emails.count} users"
puts "Pricing update failed for #{pricing_update_failed_emails.count} users"
```

emmabckstrm (commented):
@nanditagoclimate Looks good! I just had one comment. I can't approve since it's my PR originally 😅

lib/tasks/update_price_for_existing_subscribers.rake
nanditagoclimate: Agreed! Looks neat.

nanditagoclimate:
[Pricing update for existing subscribers](https://trello.com/c/weNGlIlJ/777-pricing-update-for-existing-subscribers)

lib/tasks/update_price_for_existing_subscribers.rake
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+task update_price_for_existing_consumers: :environment do
+  User.left_outer_joins(:price_increase_confirmation)
+      .select('users.*', 'price_increase_confirmations.accepted AS price_increase_response')
```
stefanlindbohm, line 5: The way I read this, it should work. But the more Railsy way would be to `.includes` the association on the query and use the association as an object, too. So like this:

```suggestion
  User.includes(:price_increase_confirmation)
    ...
```

You decide if you want to change it or not :).

lib/tasks/update_price_for_existing_subscribers.rake
```diff
@@ -0,0 +1,26 @@
...
+      .find_in_batches do |group|
+    group.each do |user|
+      next unless user.active_subscription? &&
+                  user.current_plan.retrieve_or_create_stripe_plan.product != 'offsetting_subscription_2022'
+
+      manager = Subscriptions::StripeSubscriptionManager.new(user)
+      if user.price_increase_response
```
stefanlindbohm, line 12: With the above comment, this would become:

```suggestion
      if user.price_increase_response&.accepted?
```

stefanlindbohm (approved):

lib/tasks/update_price_for_existing_subscribers.rake
nanditagoclimate: Ooh this is neat, thanks!

lib/tasks/update_price_for_existing_subscribers.rake
```diff
@@ -0,0 +1,25 @@
...
+
+task update_price_for_existing_consumers: :environment do
+  User.includes(:price_increase_confirmation)
+      .find_in_batches do |group|
+    group.each do |user|
+      next unless user.active_subscription? &&
+                  user.current_plan.retrieve_or_create_stripe_plan.product != 'offsetting_subscription_2022'
+
+      manager = Subscriptions::StripeSubscriptionManager.new(user)
+      if user.price_increase_confirmation&.accepted?
```
stefanlindbohm, line 11: Silence should be a yes

lib/tasks/update_price_for_existing_subscribers.rake
```diff
@@ -0,0 +1,25 @@
...
+  User.includes(:price_increase_confirmation)
+      .find_in_batches do |group|
+    group.each do |user|
+      next unless user.active_subscription? &&
+                  user.current_plan.retrieve_or_create_stripe_plan.product != 'offsetting_subscription_2022'
+
+      manager = Subscriptions::StripeSubscriptionManager.new(user)
+      if user.price_increase_confirmation.nil? || user.price_increase_confirmation.accepted?
+        begin
+          manager.update(Subscriptions::Plan.for_footprint(user.current_plan.footprint, user.subscription_currency)
```
stefanlindbohm, line 13: @nanditagoclimate We need to changes this to divide it by 2 after the decision to half everyone's subscription to be in line with the removal of the 2X buffer factor.

PR #625, opened 2021-12-07
Add confirmation on price increase page

emmabckstrm:
## TODO

- [x] Update with new copy
- [x] Update with new date
- [x] Update with Swedish copy

Screenshot 2022-01-10 at 3 00 53 PM

Screenshot 2022-01-17 at 4 27 31 PM

Screenshot 2022-01-10 at 3 00 26 PM

Screenshot 2022-01-10 at 3 00 05 PM


app/views/price_increase_confirmation/thank_you.html.erb
```diff
@@ -0,0 +1,16 @@
+<%= render "shared/header" %>
+
+<section class="section-padding flex flex-row">
+  <div class="max-w-3xl space-y-3">
+    <h1 class="heading-lg">Thank you!</h1>
+    <% if @accepted %>
+      <p>Your new monthly fee will be <%= current_user.subscription_price * 1.5 %> from July 1st 2021.</p>
```
emmabckstrm, line 7: This "calculation" is incorrect! Sync up with price calculation in `new.html.erb`

nanditagoclimate: *2

kallenilver:
@nanditagoclimate ok, let's wait on what we decide on in the price increase meeting tomorrow to see if we should also remove the 2x offset margin.

stefanlindbohm:
Now updated with the latest copy. Looks like this:

Price increase EN
Price increase EN confirmation
Price increase SV
Price increase SV confirmation

app/controllers/price_increase_confirmation_controller.rb
```diff
@@ -1,6 +1,41 @@
...
+  def thank_you
+    @accepted = PriceIncreaseConfirmation.where(user_id: current_user.id).first.accepted
+  end
+
+  private
+
+  def set_new_subscription
+    # This assumes the pricing in `greenhous_gases.rb` has been updated!
+    current_plan = current_user.current_plan
+    @new_subscription = Subscriptions::Plan.for_footprint(current_plan.footprint, current_user.subscription_currency)
```
stefanlindbohm, line 35: This needs to be updated with the result of our discussion on buffer factor for updating subscriptions.

PR #626, opened 2021-12-07
Revert "PDFs with Grover"

emmabckstrm:
Reverts goclimate/goclimate#619


emmabckstrm:
Revert no longer needed as the issue was solved

PR #628, opened 2021-12-07
Update about us - links and title

nanditagoclimate:


config/locales/en.yml
```diff
@@ -63,9 +63,9 @@ en:
...
+  about_us_how_do_we_stop_climate_change: How do we slow down climate change?
+  about_us_how_do_we_stop_climate_change_text: To slow down climate change, as many people as possible need to live climate-friendly lives. We believe that the best way to do this is to reduce our own climate footprint as much as possible, while balancing our emissions by financing effective climate projects that accelerate the transition to a sustainable society. The climate projects we support are all Gold Standard certified, to ensure our joint efforts are as safe and sustainable as possible. Further info about the projects we support can be found <a href="https://www.goclimate.com/climate-projects" target="_blank" rel="noopener">here</a>. We are convinced that the climate crisis can be slowed down - if there are enough of us who act and if we start NOW.
```
emmabckstrm, line 68: For the link, I would keep the actual url in the view part. Also, since it's a link to our own page, we don't need the target or rel attributes. So here it would only be `<a>here</a>`

app/views/about/show.html.erb
```diff
@@ -174,7 +174,7 @@
         <h2 class="heading-lg">
           <%=t 'about_us_how_do_we_stop_climate_change' %>
         </h2>
-        <p><%=t 'about_us_how_do_we_stop_climate_change_text' %></p>
+        <p><%= t('about_us_how_do_we_stop_climate_change_text').gsub('<a', '<a class="link" rel="noopener noreferrer"').html_safe %></p>
```
emmabckstrm, line 177: As I'm mentioning below, we don't need the rel attribute. Here I would add the url as well as the class (so that we don't have paths hard coded in our code, they might break in future). 

I would have

```
.gsub("<a", "<a class='link' href='#{the_correct_path_name_from_routes}'")
```

config/locales/sv.yml
```diff
@@ -65,7 +65,7 @@ sv:
...
+  about_us_how_do_we_stop_climate_change_text: För att sakta ner klimatförändringarna behöver så många som möjligt leva klimatvänliga liv. Vi tror att det bästa sättet är att sätta press på regeringar och världsledare samtidigt som vi minskar våra egna klimatavtryck så mycket som möjligt och balanserar våra utsläpp genom att finansiera effektiva klimatprojekt som påskyndar omställningen till en hållbar värld. De klimatprojekt vi stöttar har den säkraste och bästa av certifieringar, Gold Standard. Vi har alltid klimaträttvisa i åtanke och arbetar med 100% transparens, all info om projekten vi stöttar finns <a href="https://www.goclimate.com/se/climate-projects" target="_blank" rel="noopener">här</a>. Vi är övertygade om att den globala uppvärmningen kan saktas ner – om vi är tillräckligt många som agerar och om vi börjar NU.
```
emmabckstrm, line 68: same thing here with the link, do `<a>här</a>` as suggested in my previous comments.

emmabckstrm (approved):
:tada:

PR #629, opened 2021-12-08
Use 'fight climate change' instead of 'stop climate change'

emmabckstrm:


emmabckstrm:
[Update copy on blog](https://trello.com/c/lFxJ3NTY/774-update-copy-on-blog)

PR #632, opened 2021-12-13
Correct url helpers in sitemap controller

emmabckstrm:


emmabckstrm:
@kallenilver I guess this is a hotfix, and maybe we should fix this properly at some point? It does not currently represent all our urls.

kallenilver (approved):

PR #633, opened 2021-12-14
Add admin view for Climate reports

emmabckstrm:
Also made it possible to edit Climate reports - re-using the form view for the creation of a climate report by customers. This approach was chosen to be time efficient as this might change in the future and to keep the climate report form in one place for consistency.

Also renamed the Climate report (built spring 2021) to Data report, and moved views and controller into the Data collection namespace to avoid naming conflicts.


nanditagoclimate (approved):
🎉

PR #636, opened 2021-12-17
Add search function to Climate reports admin view

emmabckstrm:


app/models/climate_report.rb
```diff
@@ -15,6 +15,17 @@ class ClimateReport < ApplicationRecord
...
 
+  def self.search(query, limit = 100)
+    return none if query.blank?
+    return none if query.length < 3
+
+    where('contact_email ILIKE ?', "%#{sanitize_sql_like(query)}%").or(
+      where('company_name ILIKE ?', "%#{sanitize_sql_like(query)}%").or(
+        where('calculation_period ILIKE ?', "%#{sanitize_sql_like(query)}%")
+      )
+    ).limit(limit)
```
emmabckstrm, line 26: Using `ILIKE` for case-insensiveness, but it's slower as it's not using indexes if I understand it correctly. I think we might get away with this solution for now though, as we currently don't have more than 1200 records. But happy to make a simple other fix if there is a better one that I have missed

nanditagoclimate (approved):

PR #638, opened 2021-12-28
Draft: admin tools for climate report

nanditagoclimate:
- APIs for admins to be able to modify factors
- climate constants moved from yml to database
- Admin can alter formula for each emission factor

-----
What this PR doesn't do:
Handle scope specific value.

A bunch of test cases are failing, I am yet to have a look at that.


stefanlindbohm (commented):
This is a bit different of a direction than I had in mind. Not to say one is better than the other. Let's discuss after standup tomorrow, if that works for you?

The very brief summary of what I was thinking is to use formulas (through Dentaku) more heavily similar to the lifestyle calculator. But talking about this will be more effective I believe so let's continue on a call :).

nanditagoclimate:
Yes, I felt a little lost after making some progress. With a few pointers i'll be able to improve this. Thank you! We could sync after standup tomorrow.

nanditagoclimate:
Closing this PR. Will refer to this when we pick this up next.

PR #639, opened 2021-12-30
Add optional phone number in climate report

nanditagoclimate:
This gem is really cool. It is built on Google lib phone library. 
It validates all international numbers.
I have also added an initialiser and set the default country to Sweden so local customers do not have to enter their full phone number like(+46705476300) and only "705476300" will be valid and stored.


kallenilver (approved):
Looks good! Maybe change the text "you need to enter your e-mail address." to "you need to enter your contact information."?

nanditagoclimate:
Please review the Swedish and German copy as well @kallenilver  :)

kallenilver (approved):
🥳

PR #641, opened 2022-01-11
Use new product ID to make sure plans allocate right amounts of co2e

stefanlindbohm:
Plans in Stripe have the amount of offsetting set as metadata, which means that any change in price needs a new product with new plans to keep bookkeeping of offsetting allocation correct. This adds a new product ID for our new price level.

Migration of data that have been using mismatching plans can be done manually to fix what needs to be fixed.


app/models/subscriptions/plan.rb
```diff
@@ -64,7 +66,15 @@ def initialize(footprint, price, monthly_offset, stripe_plan = nil)
     end
 
     def retrieve_or_create_stripe_plan
-      @stripe_plan || Stripe::Plan.retrieve(stripe_plan_id)
+      return @stripe_plan if @stripe_plan.present?
+
+      plan = Stripe::Plan.retrieve(stripe_plan_id)
+
+      raise InconsistentStripePlanError, <<~TEXT if plan.metadata['co2e'].to_i != monthly_offset.co2e
```
nanditagoclimate, line 73: 💯

nanditagoclimate (approved):

app/models/subscriptions/plan.rb
```diff
@@ -2,8 +2,10 @@
 
 module Subscriptions
   class Plan
+    class InconsistentStripePlanError < StandardError; end
+
     BUFFER_FACTOR = 2
-    SUBSCRIPTION_PRODUCT_ID = 'offsetting_subscription_2021' # Created manually in Stripe
+    SUBSCRIPTION_PRODUCT_ID = 'offsetting_subscription_2022' # Created manually in Stripe
```
nanditagoclimate, line 8: Looks like the plan is auto created in stripe so we can remove the comment?

PR #642, opened 2022-01-12
Remove pricing buffer factor

nanditagoclimate:


nanditagoclimate:
[Remove buffer factor from subscriptions and gift cards](https://trello.com/c/jbiuEnMs/788-remove-buffer-factor-from-subscriptions-and-gift-cards)

PR #652, opened 2022-02-07
Partnership Page[BE]

nanditagoclimate:
API docs: https://drive.google.com/drive/folders/12PmG6ZE5GXRp36UgfXNYv3gD8Oh1_Zov?usp=sharing

Post deployment steps: (when required) ex:
`rake "add_organization_and_update_users[goclimate,goclimate.com,/partners/goclimate.png]"`


app/controllers/application_controller.rb
```diff
@@ -11,6 +11,11 @@ class ApplicationController < ActionController::Base
   helper_method :current_region, :canonical_url, :experiment_active?, :visitor_country
   helper ReactComponentHelper
 
+  module Status
```
kallenilver, line 14: I don't fully understand, is this used somewhere?

app/controllers/partners_controller.rb
```diff
@@ -12,4 +12,59 @@ def inshapetravel
...
+    # Organization should be a partner
+    # Logged in user should be an employee of the organization
+
+    show_page = @org&.partner? && @org_employees&.ids&.include?(current_user.id)
+    render_not_found && return unless show_page
+
+    @climate_actions_taken = UserClimateAction.where(user_id: @org_employees.ids).count
+    @co2_tonnes_avoided = @org.total_co2_avoided
+
+    @hero_top_list = hero_top_list(@org_employees.ids)
```
kallenilver, line 31: We have changed the word hero to friend on the user dashboard. I would vote for doing the change here as well, both in text and in code.

app/controllers/partners_controller.rb
```diff
@@ -12,4 +12,59 @@ def inshapetravel
...
+      .where('user_climate_actions.user_id = ? or user_climate_actions is null', user_id)
+      .select('climate_actions.id as id, climate_actions.name,
+        climate_actions.description, climate_actions.image_url,
+        user_climate_actions.id as user_accepted_action, user_climate_actions.status,
+        climate_actions.climate_action_category_id as category_id')
+      .order(user_accepted_action: :desc)
+      .map(&:attributes)
+      .each { |action| action['user_accepted_action'] = action['user_accepted_action'].present? ? true : false }
+      .each { |action| action['category_name'] = ClimateActionCategory.find(action['category_id']).name }
+  end
```
kallenilver, line 69: I guess this is not needed if we won't show this on the page (due to the react code being to inflexible)? We can talk more about this in the standup tomorrow maybe?

app/models/organization.rb
```diff
@@ -20,6 +20,14 @@ def subscription?
     climate_reports.order(created_at: :desc).first.invoice&.subscription?
   end
 
+  def total_co2_avoided
+    @total_co2_avoided = 0
+    users.each do |user|
+      @total_co2_avoided += user.total_subscription_offsetting.tonnes.round(1)
+    end
+    @total_co2_avoided
```
kallenilver, line 28: reminder: add the tons avoided by the 10 Euro per action addition, so something like: climate_actions.count * 10 / PRICE_PER_TONNE_IN_EURO

db/migrate/20220203131030_add_partner_to_organization.rb
```diff
@@ -0,0 +1,8 @@
+class AddPartnerToOrganization < ActiveRecord::Migration[6.0]
+  def change
+    add_column :organizations, :partner, :boolean, default: false
+  	add_column :organizations, :key, :string, nil: false
```
kallenilver, line 4: inconsistent indentation

db/schema.rb
```diff
@@ -61,6 +61,19 @@
...
+    t.decimal "value"
+    t.string "unit"
+    t.string "country"
+    t.string "category"
+    t.integer "scope"
+    t.text "formula"
+    t.boolean "default"
+    t.datetime "created_at", precision: 6, null: false
+    t.datetime "updated_at", precision: 6, null: false
+  end
```
kallenilver, line 75: I'm not sure what this is?

app/controllers/application_controller.rb
nanditagoclimate: This is a standard way to sending status in an API call.
I had added these when we decided to send a raw JSON in the APIs but now it's been decided to use an erb and have a react component defined there. So, I shall remove these.

app/models/organization.rb
```diff
@@ -33,4 +37,24 @@ def add_admin_from_email(email)
...
+  end
+
+  def actions_offsetted_co2
+    actions_taken = UserClimateAction.where(user_id: users.ids).count
+    cost_per_action = 10 # 10 EUR
+    currency = Currency::EUR
+    cost_per_tonne = GreenhouseGases::BUSINESS_PRICE_PER_TONNE[currency]
+
+    total_contribution = (actions_taken * cost_per_action)
+    ((total_contribution / cost_per_tonne.subunit_amount.to_f) * 100.00)
```
nanditagoclimate, line 58: @kallenilver  could you review this part please?

kallenilver: Looks good I think!

axelsson (approved):
Looks good as far as im able to understand ruby/rails and im definitely leaving the co2-calculations to Kalle 🙈 
Just an old file left that can be removed, `app/views/partners/show.html.erb`.

Good work! 🎉

app/controllers/organization_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    # Organization should exist
+    # Organization should be a partner
+    # Logged in user should be an employee of the organization
+
+    show_page = @org&.partner? && @org_employees&.ids&.include?(current_user.id)
+    render_not_found && return unless show_page
+
+    @climate_actions_taken = UserClimateAction.where(user_id: @org_employees.ids).count
+    @co2_tonnes_avoided = @org.total_co2_avoided
```
kallenilver, line 17: Could this be a `GreenhouseGases.new` object instead? It makes it a easier to print in a nice way with an own implementation of .to_s

app/models/organization.rb
```diff
@@ -20,6 +20,10 @@ def subscription?
     climate_reports.order(created_at: :desc).first.invoice&.subscription?
   end
 
+  def total_co2_avoided
+    (employees_offsetted_co2 + actions_offsetted_co2).round(2).to_f
```
kallenilver, line 24: The `round(2).to_f` is probably not needed if we use `GreenhouseGases.new` instead?

kallenilver (approved):
Good job!

Some comments that using the GreenhouseGases-model might simplify the code, but take the decision if you think this is useful or not.

🥇

PR #653, opened 2022-02-07
update action of the month

nanditagoclimate:


kallenilver (approved):
There is some code in the climate actions controller regarding action of the month as well, maybe remove that also?

kallenilver (approved):
Looks great! 🥳

PR #660, opened 2022-02-28
remove remains from action of the month

axelsson:
Removed remains from action of the month. Not sure how to proceed with the migrations to remove the column for action of the month so I'll need help with that, post or pre deploy I guess.


axelsson:
[Remove remains from action of the month](https://trello.com/c/eBQj03ZR/817-remove-remains-from-action-of-the-month)

nanditagoclimate:
PR looks good!
Great work! 🎉 

However, I am not too keen on removing the database column as when we decide to re introduce this feature, we should be able to simply cherry pick commit and create a PR and the above migration won't work with this technique.
So, I'd suggest we keep the database column as is. But I'd leave it to you. :)

nanditagoclimate (approved):

db/migrate/20220228111111_remove_action_of_the_month_from_climate_actions.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveActionOfTheMonthFromClimateActions < ActiveRecord::Migration[6.0]
+  def change
+    remove_column :climate_actions, :action_of_the_month, :boolean
```
nanditagoclimate, line 3: check if it's reversible.

PR #661, opened 2022-02-28
Allows us to publish a GoClimate app on the fortnox app store

kallenilver:
I did a quick first iteration of an integration with fortnox, so that we can publish an app in their app store and offer our accounting engine services there. 

This flow basically just allows a fortnox user to give us access to their bookkeeping, and saves the auth tokens in our back end, so that we in a later stage can get back to the customers with a price estimate.

Here's the flow:

0. Enter the fortnox-keys in the .env file

1. Activate the goclimate app in the fortnox app store: 
https://apps5.fortnox.se/app/526322bce4264c1bac1a8172ddccb7a9/integration-partner/?3819474952065185
Skärmavbild 2022-03-02 kl  15 38 19

2. change the url after pushing activate from goclimate.com/se/fortnox to localhost:3000/se/fortnox and click activate. 
Skärmavbild 2022-03-02 kl  15 41 15

3. authenticate the app through fortnox4. 
Skärmavbild 2022-03-02 kl  15 41 41

4. change goclimate.com part of the url to localhost:3000, enter form and click contact me
Skärmavbild 2022-03-02 kl  15 42 58

5. see that a `FortnoxCustomer` entry is added with auth and refresh tokens


app/controllers/fortnox_customers_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  def new
+    @code = params[:code]
+    @fortnox_customer = FortnoxCustomer.new
+  end
+
+  def create
+    fortnox_api = FortnoxApiAdapter.new
+    fortnox_params = fortnox_api.connect(params['code'])
```
axelsson, line 22: Not sure how we're usually working with logging/error handling but i accidentaly tested the flow without having the env-variables added and no error whatsoever was thrown; the row in the database was added without access/refreshtoken and the user got to the "done"-page. Might be good to have some sort of handling so that we and the users notices if something's wrong with the api-connection.

app/views/fortnox_customers/_form.html.erb
```diff
@@ -0,0 +1,31 @@
...
+          <li><%= message %></li>
+        <% end %>
+      </ul>
+    </div>
+  <% end %>
+
+  <%= hidden_field_tag 'code', @code %>
+
+  <div class="space-y-4">
+        <div class="t:flex justify-between">
```
axelsson, line 16: indentation

db/migrate/20220226205735_create_fortnox_customers.rb
```diff
@@ -0,0 +1,14 @@
+class CreateFortnoxCustomers < ActiveRecord::Migration[6.0]
+  def change
+    create_table :fortnox_customers do |t|
+      t.string :company_name
```
axelsson, line 4: company name doesnt seem to be used/set?

app/views/fortnox_customers/_form.html.erb
```diff
@@ -0,0 +1,31 @@
...
+          <%= form.label :customer_email, 'E-post', class: 'font-bold leading-10' %>
+          <div class="t:w-2/3"><%= form.text_field :customer_email, class: 'input w-full' %></div>
+        </div>
+        <div class="t:flex justify-between">
+          <%= form.label :customer_phone, 'Telefonnummer', class: 'font-bold leading-10' %>
+          <div class="t:w-2/3"><%= form.text_field :customer_phone, class: 'input w-full' %></div>
+        </div>
+        <button type="submit" class="button button-cta w-full">Skicka in! (steg 2 av 2)</button>
+      </div>
```
axelsson, line 29: Shouldn't these fields be required? Now you can successfully submit the form without any data at all.

axelsson (changes requested):
Unsure how much of a poc this is so just ignore the comments that you dont feel are relevant at this stage.

app/controllers/fortnox_customers_controller.rb
kallenilver: Great thinking, I'm a bit too fast sometimes :) I've added the not null to the api-columns so that we will throw an exception if we don't get them from the api. That way at least we will be notified if something is up.

db/migrate/20220226205735_create_fortnox_customers.rb
kallenilver: Good catch, removed it now!

app/views/fortnox_customers/_form.html.erb
kallenilver: Good thinking, I made name and email html5 required now. Phone is ok to skip I think.

axelsson (approved):
Good work, let's see if fortnox wanna try it out! =)

PR #664, opened 2022-03-14
Upgrade tailwind

axelsson:
* Upgraded tailwind v2 -> v3. Some names of tailwind classes have been updated.
* Removed a few deprecated screen sizes (sm,md,lg) and replaced with the current correct ones.
* Had to change the postcss config formatting, I compared the size of the packages when using prod-config and it seems to be the same so I guess it works as intended.

It will be easier to look at the changes when hiding whitespaces cause the .css-components are from now on wrapped in `@layer components { ... }`.

Test by clicking around and just making sure it looks and behaves as before.


axelsson:
[Upgrade tailwind v2 -> v3](https://trello.com/c/xAE5V8Fu/822-upgrade-tailwind-v2-v3)

kallenilver (changes requested):
For some reason the actions are really narrow now in upgrade-tailwind compared to master.
Skärmavbild 2022-03-15 kl  08 36 02

PR #671, opened 2022-03-22
Fix eslint for react and refactor

axelsson:
* The separate eslint-file for react is now merged into the real one in the root which is used when running `bin/yarn lint`
* Removed 'prettier' since I didn't see any need for it
* Removed some unused code in climate actions
* Fixed all linting errors

I don't think there's much use of looking through all changes here except for package.json and the eslint config but I'd be glad if you ran the code locally and tested a few of the features (especially climate actions) for a short time. I've tried to find everything to test but it's not impossible that I've missed a few things. :)


axelsson:
[Fix eslint-setup for react components](https://trello.com/c/IU1Xn9n1/829-fix-eslint-setup-for-react-components)

PR #674, opened 2022-03-29
Annoucements and preview

axelsson:
These changes are primarily made for the upcoming partnership campaigns, to make it easier for visitors to understand what we do and sign up.

- Added "announcement" to landing page for displaying short messages about news. When changing this to a new announcement the version number in the cookie has to be manually increased in the code. I figured this was the easiest solution while we evaluate the need for this. This particular one is only targeting logged in users but that check can easily be removed if the next one is supposed to target all users.
- Added a preview in the sign up process to give a sneak peak of the dashboard.
- Removed the option to sign up while paying for several people, in order to de-clutter the sign up process. I chose to only comment out the code cause there was so much logic surrounding it and I'm not sure it should be entirely deleted.
- Had to make a few adjustments to the modal to make it work for larger content (especially on mobile).
- Decreased size of organization logo.
- Removed 'monthly action' leftover in admin.

If you have any styling suggestions to make it prettier they're welcome, I've improvised. :)


axelsson:
[Update site for partnership campaigns](https://trello.com/c/ydjUuxgZ/833-update-site-for-partnership-campaigns)

app/assets/main/controllers/registrations/membership_choice_controller.js
```diff
@@ -37,7 +37,7 @@ export default class RegistrationsMembershipChoiceController extends Controller
       swapToInactiveClassList(this.stripeCardElementTarget);
       this.showNoSubscriptionInfo();
     } else {
-      this.registrationPriceController.update();
+      // this.registrationPriceController.update();
```
kallenilver, line 40: I guess we can simply remove this? Or is there a reason we only commented it out?

app/views/devise/registrations/_sign_up_form.erb
```diff
@@ -21,7 +48,9 @@
             <input class="shrink-0 mr-2" type="radio" name="membership" id="single" value="single" checked data-target="registrations--membership-choice.choice" data-action="click->registrations--membership-choice#handleChange">
             <span><%= sanitize(t('views.registrations.membership.single')).gsub('<span', '<span class="font-bold"').html_safe %></span>
           </label>
-          <label class="flex flex-row items-center p-3 bg-gray-pastel rounded cursor-pointer"
+          <%
```
kallenilver, line 51: Green feedback (as in not very important to me at all): Guess I prefer to remove it here, the code will be in git if we need it again

kallenilver (approved):
Looks good! 🥳

app/views/devise/registrations/_sign_up_form.erb
axelsson: I'll make a separate PR for this if it's okay. There was so much logic surrounding it i didnt want to clutter this PR.

kallenilver: Sounds good!

PR #676, opened 2022-04-05
Move business dashboard to org

axelsson:
* Moved and renamed the business dashboard from dashboard controller to organization
* Updated routing correspondingly, please take an extra look at those lines, I'm not confident I solved it in the best the way :)
* Renamed current organization dashboard (netlight...) to employee dashboard
* Moved all related translations
* Added a sidebar to the organization dashboard
* Added placeholder pages for "climate actions" and "projects" under organization. We'll see what ends up on those pages, maybe they'll need to be renamed and the routes changed when we decide exactly what content they are gonna hold.

Since the dashboard is not yet used live I don't see any harm in merging this into master but if you prefer it we can keep it on a separate branch.

The only question left is whether we're fully live with the netlight dashboard, in that case the routing will give us a headache I think.

One thing I also noticed that I thought was a bit strange is that the organization in the dashboard is always picked up from the user rather than the url so that might give us some strange routing problems, since I guess any name in organization/<name>/dashboard would lead to the users organization. So I guess that's something we should fix, to make it pick up the org from the url rather than the user?


axelsson:
[Navigation and empty templates for climate actions and project/offset pages, move business dashboard into organization](https://trello.com/c/OqJkx3Nr/842-navigation-and-empty-templates-for-climate-actions-and-project-offset-pages-move-business-dashboard-into-organization)

config/routes.rb
```diff
@@ -130,7 +130,17 @@
     get 'partners/bokanerja'
     get 'partners/inshapetravel'
     get 'partners/flygcity'
-    resources :organization, controller: 'organization', only: [:show], param: :key
+
+    # Organization
+    resources :organization, controller: 'organization', only: [], param: :key do
+      member do
+        get 'dashboard', to: 'organization#dashboard'
```
nanditagoclimate, line 137: Doesn't this get auto-resolved? Any reason why we need to be specific?

config/routes.rb
```diff
@@ -130,7 +130,17 @@
...
+        get 'subscribe', to: 'organization#subscribe'
+        get 'dashboard/climate-actions', to: 'organization#climate_actions'
+        get 'dashboard/projects', to: 'organization#projects'
+        get 'employee-dashboard', to: 'organization#employee_dashboard'
```
nanditagoclimate, line 141: I would personally avoid hyphens "-" and use underscore "_" or even camelCase if needed. But "_" is better.  (PS: "_" is not an emoji) :D

nanditagoclimate (approved):
Please take a look at the routes again else good job on the PR. 🎉

config/routes.rb
axelsson: Doesnt seem like it, if i remove that row it doesnt work 🤷‍♀️

config/routes.rb
axelsson: We're using hyphens everywhere else afaik and i think it's more of a standard for urls than the other suggestions

axelsson:
@kallenilver I suppose I should add the redirect of https://www.goclimate.com/organization/<org> -> https://www.goclimate.com/organization/<org>/employee-dashboard in this PR before we merge, right?

config/routes.rb
nanditagoclimate: My bad, I was confusing them with something else.

config/routes.rb
nanditagoclimate: Ah, I see. You could either do what you're doing or define a `path`. Refer to line # 90 in routes file.

kallenilver:
> @kallenilver I suppose I should add the redirect of https://www.goclimate.com/organization/ -> https://www.goclimate.com/organization//employee-dashboard in this PR before we merge, right?

Do you mean from https://www.goclimate.com/organization/{org_name} to https://www.goclimate.com/organization/{org_name}/employee-dashboard then yes :)

axelsson:
@kallenilver @nanditagoclimate I updated the routes now in the last commit, please take a second look :)

kallenilver (approved):
This looks awesome! Onwards and upwards!

PR #677, opened 2022-04-08
More business calc translations

axelsson:
Now shows correct heating type for when selecting gb as country. Started with adding country versions to the "heating"-translation key but realized it was used in lots of other places too, where it would be too much work to handle the different countries so I solved it by adding a new key instead.

Also fixed some rubocop-warnings.


axelsson:
[Fix "heating" translation in business calculator](https://trello.com/c/8mOt5LQN/847-fix-heating-translation-in-business-calculator)

app/controllers/admin/data_collection/forms_controller.rb
```diff
@@ -88,7 +88,7 @@ def set_units
       end
 
       def verify_editable
-        redirect_to admin_forms_url, notice: 'Published/archived form can\'t be edited.' unless @form.draft? # rubocop:disable Metrics/LineLength
+        redirect_to admin_forms_url, notice: 'Published/archived form can\'t be edited.' unless @form.draft?
```
kallenilver, line 91: not your code and very unimportant, but I prefer `"` to avoid escaping `'` within `'`

kallenilver (approved):
🔥

app/controllers/admin/data_collection/forms_controller.rb
axelsson: i agree, easily changed

PR #682, opened 2022-05-03
Add business calculator support for Germany and USA

nanditagoclimate:


axelsson (changes requested):
I pushed a commit to add a swedish translation for the tooltip regarding what countries we support and also added logic to use sqft and miles for the us (as well as uk). When I perform the calculation for se/de/us setting all fields to the value 100 I get exactly the same results when I submit the form. I expected the results to differ since we're supposed to use different values for different countries but apparently they do not. Have I misunderstood or are we not actually using those numbers?

PR #683, opened 2022-05-10
Add scope and compare graphs to accounting report pdf

kallenilver:


app/models/business/accounting_emissions_calculation.rb
```diff
@@ -106,6 +121,35 @@ def emission_percentages
...
 
+    def scope_2_emissions
+      SCOPE_EMISSIONS.inject(0) do |sum, (emission_name, scope)|
+        scope == 2 ? sum + send(emission_name) : sum
+      end
+    end
+
+    def scope_3_emissions
+      SCOPE_EMISSIONS.inject(0) do |sum, (emission_name, scope)|
+        value = emission_name == :other_emissions ? other : send(emission_name)
```
nanditagoclimate, line 132: this value can be 0 if user has not entered anything.

app/models/business/accounting_emissions_calculation.rb
```diff
@@ -106,6 +121,35 @@ def emission_percentages
...
+        scope == 3 ? sum + value : sum
```
nanditagoclimate, line 133: this will fail with a nil comparison so we can change line 32 to 
value = emission_name == :other_emissions ? other : send(emission_name) || 0

nanditagoclimate (changes requested):

nanditagoclimate (approved):

PR #684, opened 2022-05-11
add rack cors for handling preflight requests in API, remove older co…

axelsson:
…rs handling

Added CORS handling in the API to allow browsers to make preflight requests. The current code for the API works fine if you use curl/make the request from a server/postman etc but not when making them from a browser, since the browser adds additional security. I think Stefan tried to solve this earlier in this commit (https://github.com/goclimate/goclimate/commit/3821845f34b582ef8ccff79d2bdec426023d370b) but it seems it didn't work correctly. 

I also added handling for requests to our upcoming endpoints at `/api/v1/user/` where we'll need to include credentials (session cookie) in the request. That endpoint only exists in the branch `user-api-mock` currently.

Tested locally and I had to update the tests; it seems preflight requests are a special case of a OPTIONS request.

I'm not very confident about the logic for CORS in prod though since the API is located under a subdomain. As I've understood it it's no catastrophe if the API is unavailable for a few minutes so I'm thinking we'll just have to deploy and test it. Any additional ideas of how to test that are very welcome.

Since the new endpoint does not yet exist, the main concern of this PR is just to ensure that it all works as before 😅


axelsson:
[CORS handling in API](https://trello.com/c/eQuAs0bT/877-cors-handling-in-api)

PR #686, opened 2022-05-12
Upgrade gems and rubocop version in code climate

nanditagoclimate:
- Upgraded gem versions through bundle update
- Upgraded rubocop version to support ruby 2.6
- Changed the rubocop channel on code climate so it is in sync with our local environment.

Good practices for future:
1. Run bundle update regularly or every time we add a gem and commit it
2. Change rubocop channel on codeclimate config file if we upgrade it

All our local and remote needs to have the same gem versions at all times.


nanditagoclimate:
https://docs.codeclimate.com/docs/rubocop Doc for reference.

kallenilver (approved):
Great! Lots of gem updates, but as long as the tests pass I guess we should be fine :)

PR #688, opened 2022-05-16
make current projects clickable, link leads to blog post

axelsson:
According to Adreas suggestion (which I very much agree with) the "Current projects" are now clickable and the link leads to the corresponding blog post.


kallenilver (approved):
Level up!

PR #689, opened 2022-05-16
User api endpoint

nanditagoclimate:
APIs - https://docs.google.com/document/d/1Yxu9UyQzFU_vUlBI1W-cWzvHPZCVw4oJ12BprC835-A/edit
As discussed, we could go live with this and test how the app and user authentication works on production.
Let's continue building this in a new branch.


app/models/organization.rb
```diff
@@ -40,6 +42,13 @@ def add_admin_from_email(email)
 
   private
 
+  def associate_with_users
```
kallenilver, line 45: I'm not sure why we should do this? Every time we add an organisation the users with the same domain name as the org becomes part of the org?

app/serializers/api/v1/overview_serializer.rb
```diff
@@ -0,0 +1,109 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class OverviewSerializer < ActiveModel::Serializer
+      EMISSION_PER_HAMBURGER = 2.2 # kg CO2
```
kallenilver, line 6: Should we place these values in the serializer?

kallenilver (approved):
Looks good! I like the `active_model_serializers` thing!

app/serializers/api/v1/overview_serializer.rb
nanditagoclimate: Valid point. I shall move it to yml!

app/models/organization.rb
nanditagoclimate: Yes! You're right.

app/controllers/api/v1/projects_controller.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class ProjectsController < ApiController
+      before_action :authenticate_user!
+
+      def show
+        organization_id = current_user.organization&.id
+        return unless organization_id
```
axelsson, line 10: Why do we check for org id here but not in overviews for example? It feels a bit inconsistent?

app/serializers/api/v1/overview_serializer.rb
```diff
@@ -0,0 +1,109 @@
...
+
+      def latest_calculation_period_year
+        object.pluck(:calculation_period).max
+      end
+
+      def climate_report_count
+        object.count
+      end
+
+      def pretify(label)
```
axelsson, line 104: -> prettify

axelsson (commented):
I'm assuming you're gonna fix the camel case thingy before this is done so I'll just leave a few minor comments meanwhile.

app/serializers/api/v1/overview_serializer.rb
```diff
@@ -0,0 +1,109 @@
...
+
+      def scope_based_emissions_percentage
+        object.map { |cr| [cr.calculation_period, cr.calculation.even_scopes_percentages.sort.to_h] }.to_h
+      end
+
+      def scope_based_emissions_value
+        object.map do |cr|
+          [cr.calculation_period, {
+            'Scope 2': cr.calculation.scope_2_emissions,
+            'Scope 3': cr.calculation.scope_3_emissions
```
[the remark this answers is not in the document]
nanditagoclimate: Good question, Yes! Our calculator doesn't support this at this moment but we can add it throughout once we have the factors in.

PR #692, opened 2022-05-24
handle when api key is nil

nanditagoclimate:
https://sentry.io/organizations/goclimate/issues/3295828757/?project=1765570&referrer=slack


kallenilver (approved):
Should be fine!

PR #696, opened 2022-06-10
exclude public endpoint from cors policy

axelsson:
This config says "don't use cors for flight_footprint but use it for everything else located under /api/v1/*". If we enable it for _everything_ then the specified origin will be required for flight_footprint as well, which breaks it.


axelsson:
[Update cors policy](https://trello.com/c/Q0FcaXHX/916-update-cors-policy)

kallenilver (approved):
Looks good as far as I can understand! 💯

axelsson:
[Upgrade Sentry](https://trello.com/c/KkD0M5Ub/825-upgrade-sentry)

PR #697, opened 2022-06-14
More apis cat v1

nanditagoclimate:
API updated here: https://docs.google.com/document/d/1Yxu9UyQzFU_vUlBI1W-cWzvHPZCVw4oJ12BprC835-A/edit#heading=h.lvvy5jvcj0k7


axelsson (changes requested):
Nice! The reports data looks good! I don't know how to test the AccountingEmissions-thingies, how do I do that?

I saw in the API doc that "calculation_period": "year" for the AccountingEmissions data, could we get the actual period out there instead? If I understand the code correctly we use start date and end date for those instead. If it's too messy to handle now I guess we could just hide it for those until we decide it's worth fixing.

I get 500 when requesting the projects now. I've got one invoice with a project connected and another one without a project. I don't really understand what goes wrong but I've added the stack trace underneath here. Get back to me if you can't reproduce it and we'll figure out what kind of setup is needed for this :)
 
`undefined method `[]' for nil:NilClass excluded from capture: DSN not set

NoMethodError (undefined method `[]' for nil:NilClass):

app/serializers/api/v1/contributed_projects_serializer.rb:11:in `calculation_period'
app/controllers/api/v1/projects_controller.rb:18:in `show'`

nanditagoclimate:
Good point. I have added start and end date in accounting reports API! For now, the format is "2022-06-14 : 2022-06-16"
Let's discuss on a better format if needed on standup.
Regarding the projects API, the reason you see the error is because post condition check on projects name, the contributed projects returned [nil]. Now I have fixed it and you shouldn't be able to see that error anymore.

axelsson (approved):
Nice, works out just fine =)

PR #699, opened 2022-06-21
Election campaign

axelsson:
Campaign page for the swedish election, will be available at www.goclimate.com/valet
The video will be replaced when we get the newly produced one.
I don't see any problem with us making it public before the video is added, but let me know if you think otherwise. Tove has reviewed the content in the staging environment.


kallenilver (approved):
Lets go! 👍

PR #701, opened 2022-06-30
allow usage of url param 'redirect_to' on sign in

axelsson:
Changes made to not lose users when they're required to login for CAT.
This allow us to link to CAT and let users that are not logged in first get redirected to the log in page and then redirected back to the CAT url they're supposed to visit once they've logged in.

Example usage: `http://localhost:3000/users/sign_in?redirect_to=http://localhost:3001/business/goclimate`

As far as I've tested the old built in version of redirect after login (handled by session) still works with these changes.


nanditagoclimate (approved):
This is good and it'll give us a lot of flexibility in the future as well. So good job 🎉

PR #703, opened 2022-08-03
Upgrade sentry

axelsson:
`sentry-raven` is deprecated, and now replaced with `sentry-rails` and `sentry-ruby`. The sentry solution for frontend seems to have been broken for a while. Changed the solution somewhat to user their lazy loader (read more [here ](https://docs.sentry.io/platforms/javascript/install/lazy-load-sentry/) if interested) and remove the integration for tracing (I don't think we used that one anyway?). Tested in staging, js errors works, could not trigger a backend error unfortunately so that's untested.


axelsson:
[Upgrade Sentry](https://trello.com/c/KkD0M5Ub/825-upgrade-sentry)

kallenilver (approved):
Good catch. Code looks good to me!

PR #704, opened 2022-08-05
add button for starting subscription beneath footprint calc result

axelsson:
Added text + button to make users take action when they view their dashboard or complete a new calculation and do not already have a subscription. I also modified the text in the "update payment" view since it's really not an update if you haven't got a subscription already.

I improvised on both design and text so please give me feedback if you have a better idea :)


axelsson:
[Add explanation for subscription and a button for actually doing it when finishing a footprint calculation if you're already logged in without a subscription](https://trello.com/c/Zu32Vh0G/844-add-explanation-for-subscription-and-a-button-for-actually-doing-it-when-finishing-a-footprint-calculation-if-youre-already-logg)

config/locales/views/dashboard.en.yml
```diff
@@ -5,7 +5,9 @@ en:
         subscription_coverage: Your current monthly contribution is %{subscription_price} which corresponds to %{footprint_coverage}% of your yearly carbon footprint.
         previous_footprints: See previous footprints
         update_subscription: Update subscription
-        subscribe: Balance your footprint
+        help_financing: Help financing climate projects
```
kallenilver, line 8: "Help finance climate projects" sounds better in my ear?

kallenilver (approved):
Yey!

PR #706, opened 2022-08-11
Bug fix on climate report creation

nanditagoclimate:


axelsson (changes requested):
When I've created a report and go back to the list and then follow the link to "show" for a certain report (example http://localhost:3000/admin/data_collection/reports/2)  I get an error saying "No template for interactive request
Admin::DataCollection::ReportsController#show is missing a template for request formats: text/html"

PR #708, opened 2022-08-15
staging settings for testing settings for session cookie

axelsson:
Problem: CAT needs to use the session cookie, today it's not allowed since the session cookie is set to the domain `www.goclimate.com`  (see image) and cat is at app.goclimate.com.
image
Possible solution: Specify session cookie name, domain and tld (top level domain) instead of using defaults.

I needed to change the name of the cookie (from _go_climate_session -> _goclimate_session) since not changing it made rails very confused and destroyed the possibility for current users to log out. Giving it a new name seemed to work for new users, and old users could still use their old session cookie. I don't really understand why the old remaining cookie would still work though. Found this thread of another person with the same question at least: https://stackoverflow.com/questions/31593884/change-rails-session-cookie-domain-without-logging-users-out

The config that's supposed to work in real production is the one that's commented out, and the used one is only for testing on stage, so a change will be required before deploying this. I don't even know if it makes sense to test this through staging but please just take a look at it and see if it makes sense :D


axelsson:
[Figure out session sharing between subdomain and domain](https://trello.com/c/RF1LXdxd/927-figure-out-session-sharing-between-subdomain-and-domain)

config/environments/production.rb
```diff
@@ -96,4 +96,10 @@
 
   # Do not dump schema after migrations.
   config.active_record.dump_schema_after_migration = false
+
+  # Share session cookie with subdomains
+  # config.session_store :cookie_store, key: '_goclimate_session', :domain => 'goclimate.com', tld_length: 2
+
+  # This one is for testing in staging; other domain and therefore other tld length
+  config.session_store :cookie_store, key: '_goclimate_session', :domain => 'goclimateneutral-staging.herokuapp.com', tld_length: 3
```
nanditagoclimate, line 104: 👍 
Let's test this one on staging!

nanditagoclimate (approved):

config/environments/production.rb
axelsson: I've done that already but I'd be glad if you'd do that test as well! Or tell me if you wanna look at it together.

PR #709, opened 2022-08-16
Add irb config

nanditagoclimate:
Is it just me or are these suggestions caused due to ruby upgrade irritating you as well?

Screenshot 2022-08-16 at 3 21 34 PM

Here's a config change that stops these suggestions.
Upto the team to decide if we can deploy it.


kallenilver (approved):
I haven't tried it so I have no opinions, so sounds good to disable it if you want to! 😄

axelsson (approved):
Yeah I'm fine with removing them, they messed with my terminal 🤷‍♀️

PR #711, opened 2022-08-17
Send org key in api requests cat v1

nanditagoclimate:
The APIs will now accept a param called 'key' where we can pass the org key.
Refer to API doc - https://docs.google.com/document/d/1Yxu9UyQzFU_vUlBI1W-cWzvHPZCVw4oJ12BprC835-A/edit#heading=h.pg3ahgw5lf9a

We fetch organization details from the param key and send data if any of the below conditions is met:
1. logged in user is a super admin

<or>

2.logged in user is an org admin for that organization

While creating an organization through Admin tool, key is now made mandatory.


app/controllers/api/v1/api_controller.rb
```diff
@@ -15,6 +15,10 @@ def url_options
         super
       end
 
+      def set_org_for_super_admin(org_key)
```
kallenilver, line 18: Seems like change of names here could be a good idea if the method only finds an org and does not set an org?

kallenilver: Or should it set something that is missing?

kallenilver (approved):

axelsson (commented):
Since the frontend doesn't know or care for whether the user is a super admin or not, there's no need to handle "no key sent", there will always be an org key included so I think we should always find the organization by that key and ignore the org that the user is connected to.

I guess we will want the users to be org_admins as CAT will not only be read only mode and we've lot's of users connected to organizations. Seeing CAT in the menu requires admin role atm (as far as I remember).

We should probably also auth the users when they edit climate actions for businesses.

app/controllers/api/v1/business_climate_actions_controller.rb
```diff
@@ -6,9 +6,13 @@ class BusinessClimateActionsController < ApiController
       before_action :authenticate_user!
 
       def show
-        organization_id = current_user.organization&.id
+       return unless params['key']
```
axelsson, line 9: indentation is off here

app/models/organization.rb
```diff
@@ -6,7 +6,9 @@ class Organization < ApplicationRecord
   has_many :users
   has_many :user_admins
 
-  validates_presence_of :name
+  validates_presence_of :name, :key
+  validates_uniqueness_of :name, :key
```
axelsson, line 10: This prevents creation of a new org in admin without a key, is that intentional?

axelsson (commented):
Lots of code duplication, here since all affected endpoints require the org and the check for access. If it's possible to break it out into a common function somewhere that would be nice. Not sure what context would be the most suitable for that though?

app/controllers/api/v1/business_climate_actions_controller.rb
axelsson: Also I find it strange that the API returns http 204 if the key is missing or if the key belongs to a user that doesn't have access, should probably be 4XX (not sure?) and 403 for those.

app/models/organization.rb
nanditagoclimate: Yes.

nanditagoclimate:
I see your point. We have something called helpers but that's where common business logic goes but since these are just checks they need to be at controller level itself. This is also more readable. Further refactoring would be over-engineering.

axelsson (approved):
I tried this out with CAT, works smooth 👍

PR #713, opened 2022-08-23
Cat menu updates

axelsson:
I updated the menu with the new layout to match CAT. If the user has access to anything in the "business section" the menu will be split between their personal links and the business links.

I removed the current business page and all related pages.
The thing that's still left from the previous work with the business page is the two rake tasks (`lib/tasks/add_climate_report_invoice_email_as_organization_admin.rake` `lib/tasks/connect_climate_reports_to_organizations.rake`) and two emails (`spec/mailers/previews/business_mailer_preview.rb`) since we might be needing something like that with the new user flow.

The link to CAT is still limited to `current_user.organization&.key == 'goclimate'` so this should be safe to deploy.


axelsson:
[Update rails app user menu to include CAT](https://trello.com/c/7npoUh4D/948-update-rails-app-user-menu-to-include-cat)

axelsson:
[Update rails app user menu to include CAT](https://trello.com/c/7npoUh4D/948-update-rails-app-user-menu-to-include-cat)

app/views/shared/_header.html.erb
```diff
@@ -24,29 +24,42 @@
...
                       <% end %>
                       <li><%= link_to t('settings'), edit_user_registration_path , class: 'link-ui' %></li>
                       <hr class="hidden d:block d:dropdown-separator">
-                      <% if current_user.organization.present? && current_user.organization.key == 'goclimate' %>
-                        <li><%= link_to t('views.shared.header.navigation.business_dashboard'), dashboard_organization_path(current_user.organization.key) , class: 'link-ui' %></li>
+                      <% if show_business_section %>
+                        <li class="span cursor-default p-2 text-sm text-gray-shade-2"><%= current_user.organization.name %></li>
+                        <% if current_user.organization&.partner? %>
+                          <% if current_user.organization.key == 'euma' %>
+                            <li><%= link_to 'Employee dashboard', "/organization/euma/member-dashboard", class: 'link-ui' %></li>
```
kallenilver, line 53: Could we change `Employee` to `Member` here?

public/i18n/en.json
```diff
@@ -2008,7 +1993,7 @@
             "self_serve": "Self-serve climate reporting"
           },
           "navigation": {
-            "business_dashboard": "Business Dashboard",
+            "cat": "Climate Action Tool",
```
kallenilver, line 1996: I wonder if people will understand Climate Action Tool in the context of the menu, maybe it's better to say `Business Tool` to start with? But we also want to be consistent.. Maybe deploy as is and we'll see how it feels in prod?

kallenilver (approved):
A few comments, nothing stopping. Nice work! 🔥

public/i18n/en.json
axelsson: I'm thinking that by the time they've logged in with access to CAT they will have either used the current business form to sign up with, where we talk about CAT, or have gotten an email saying how they now have access to CAT so hopefully it already means something to them.

PR #714, opened 2022-08-28
CAT v1 improvements[BE]

nanditagoclimate:
This PR fixes:

- [x] Removes other categories
- [x] Links accounting emission data entered in Admin tools to  CAT
- [x] Picks latest report for a given year


nanditagoclimate:
Please do a detailed review - especially the logic part. 🤞

axelsson (changes requested):
The data needed for the comparison chart doesn't seem to be included at all? That is, the average data for other companies to compare with. Also, the "other" category was only supposed to be excluded for that comparison data. We still wanna show it everywhere else.

axelsson (commented):
Seems to work just fine, as long as "calculation period" is a number. If the customer has a fiscal year and fills the business form with "june 18 - may 19" (or anything that's not a number) the code crashes. One solution could maybe be to change the business form to include both a longer name for the report and maybe a starting year? or something that we can use to identify the year of the time period with. I guess we could merge this code anyway but we'll probably need to solve this before we go live, to prevent customers from getting linked to broken CAT pages just cause they added anything else than a number into the field.

axelsson:
> Good point. I will handle the calculation period using regex for climate reports with calculation period as 'year'. The rest of the reports we can ignore. Also, I am thinking the following changes for business calculator:
> 
> 1. dropdown for year
> 2. Introduce new field 'Reporting period starting year'
> 3. Rename Reporting period to 'Financial year' or something similar.
> 
> We will have to backfill data for point(2) for existing data and for that we can write a regex to extract the first 4 integers in the 'Reporting period' column.
> 
> I'd suggest we do this alongside CAT so the future data will be clean!

If we make those changes I don't see a need for the regex when fetching the climate reports, why would that be needed? Then we should just be able to use the new field instead. Btw for CAT we will probably still want to display the value from the name field (today's calculation period), to use when displaying the data to enable them to clarify for example that the data represents "mar 2020 - feb 2021". 🤔

nanditagoclimate:
Regex only for backfilling data. And, yes! we can continue to show the current `Reporting period` in CAT reports.

PR #718, opened 2022-09-02
Add comparison data to overview controller

nanditagoclimate:
Comparisons data will now be returned in Overview API. Please refer to the [API document](https://docs.google.com/document/d/1Yxu9UyQzFU_vUlBI1W-cWzvHPZCVw4oJ12BprC835-A/edit#).

I've also renamed category BUSINESSTRIPSEMISSIONS in AccountingEmissionsCalculation to  BUSINESSTRIPS to be uniform across Climate Reports. I've tested accounting climate report generation and other actions and I did not see an issue but please have a look!


kallenilver (approved):
I haven't tried it out, but looks good as far as I can tell!

app/controllers/api/v1/overviews_controller.rb
```diff
@@ -19,10 +19,16 @@ def show
...
+        accounting_reports = Business::AccountingEmissionsCalculation.where.not(organization_id: organization.id)
+
         overview_data = yearly_emissions.present? ? Api::V1::OverviewSerializer.new(yearly_emissions) : {}
 
+        comparisons_data = (climate_reports || accounting_reports).present? ? Api::V1::ComparisonSerializer.new(climate_reports, accounting_reports).compare_average_emissions : {}
+
         render json: {
-          overview: overview_data
+          overview: overview_data,
+          comparisons: comparisons_data
```
axelsson, line 31: Why is this placed "outside" of overview? It's still a part of the same landing page, just as the subcategories/scope data is.

app/controllers/api/v1/overviews_controller.rb
```diff
@@ -19,10 +19,16 @@ def show
 
         yearly_emissions = merge_accounting_and_climate_reports(climate_reports, accounting_reports)
 
+        climate_reports = ClimateReport.where.not(organization_id: organization.id)
+        accounting_reports = Business::AccountingEmissionsCalculation.where.not(organization_id: organization.id)
+
         overview_data = yearly_emissions.present? ? Api::V1::OverviewSerializer.new(yearly_emissions) : {}
 
+        comparisons_data = (climate_reports || accounting_reports).present? ? Api::V1::ComparisonSerializer.new(climate_reports, accounting_reports).compare_average_emissions : {}
```
axelsson, line 27: As I understand it, these climate/accounting reports are from all years (not only from the latest year that we will compare with), is this how it's done in the old comparison as well?

axelsson (commented):
A couple of questions :)

app/controllers/api/v1/overviews_controller.rb
axelsson: To elaborate on this: I'm very confused now cause I looked at two older reports and in one we show the comparison data per employee but only for the categories, and in another one we show it per subcategory as well. I'm kinda sure that we settled on only showing the main categories but apparently I've missed the fact that it's show per employee. Not really sure what to do with this.

app/controllers/api/v1/overviews_controller.rb
nanditagoclimate: Agreed. But the data needed for comparisons is different from what the rest of the overview API needs. We can think of serializers as different functions is a sense and since for comparisons we need different data points, I have separated them from overview serializer.

app/controllers/api/v1/overviews_controller.rb
nanditagoclimate: 1. Yes
2. The current comparisons shows average emissions only. If we need to show emissions per employee, I think we can do that as well but I'll need @kallenilver 's help to understand how it's done in the older report generation. I don't understand that code :|

kallenilver: @nanditagoclimate I'm very slow, this is very much my bad. I'd be happy to look at the code together! Tomorrow after standup?

nanditagoclimate: No worries! Sounds great @kallenilver .

app/controllers/api/v1/comparison_serializer.rb
```diff
@@ -5,23 +5,15 @@ module V1
...
-      end
-
-      def compare_average_emissions
-        reports_count = @climate_reports.count + @accounting_reports.count
-
-        all_categories = @climate_reports.map{ |cr| cr.calculation.categories } + @accounting_reports.map { |acr| acr.categories }.flatten
-
-        average_emissions = {}
+      def average_emissions
+        emissions = {}
```
axelsson, line 10: I tried to understand how this works so I added a `puts object` here to see what the object looked like and I got this:

```
{"energy"=>49, "business_trips"=>77, "meals"=>1, "material"=>241, "other"=>82}
{"energy"=>35, "business_trips"=>553, "meals"=>1, "material"=>266, "other"=>82}
{"energy"=>0.19699432892249527410207939508506616257e2, "business_trips"=>0.20791682419659735349716446124763705104e3, "meals"=>0.527599243856332703213610586011342155e1, "material"=>0.265695652173913043478260869565217391304e3, "other"=>111}
{"energy"=>0.852215139442231075697211155378486055777e3, "business_trips"=>0.1298693227091633466135458167330677290837e4, "meals"=>0.23836653386454183266932270916334661355e2, "material"=>0.435549800796812749003984063745019920319e3, "other"=>407}
{"energy"=>0.598135728542914171656686626746506986028e3, "business_trips"=>0.3383686626746506986027944111776447105788e4, "meals"=>0.3461477045908183632734530938123752495e2, "material"=>0.234918163672654690618762475049900199601e3, "other"=>904}
```

but the resulting object I get in the response is this 

```
    "average_emissions": {
      "energy": 310,
      "business_trips": 1104,
      "meals": 13,
      "purchased_goods": 288
    }
```

and I don't understand how those numbers line up? I would've guessed that energy would be (49+35+0.2 + 0.8 + 0.5)/5 = 17.1, what am I missing?

app/controllers/api/v1/comparison_serializer.rb
```diff
@@ -5,23 +5,15 @@ module V1
...
-        @climate_reports, @accounting_reports = climate_reports, accounting_reports
-      end
-
-      def compare_average_emissions
-        reports_count = @climate_reports.count + @accounting_reports.count
-
-        all_categories = @climate_reports.map{ |cr| cr.calculation.categories } + @accounting_reports.map { |acr| acr.categories }.flatten
-
-        average_emissions = {}
+      def average_emissions
```
axelsson, line 8: Can we name this average_emissions_per_employee?

axelsson (changes requested):
The api-response does not contain the company's emissions per category per employee (only the total) which is needed for comparing with the average values.

app/controllers/api/v1/comparison_serializer.rb
nanditagoclimate: You're right about the math! Just 1 correction here. "0.19699432892249527410207939508506616257e2"  is actually 19.6 as its e^2. Then the numbers add up!

axelsson: Oooooh I totally missed the e2 in the end!

axelsson (approved):
Nice, this works! 🎉 

A small note: `emissions_per_employee` still contains the "other" category but I can just hide that in the frontend unless you wanna change it in backend.

PR #720, opened 2022-09-02
rename session cookie and set domain to allow sharing cookie with sub…

axelsson:
…domain

⬆️ The real config for prod 🤞


nanditagoclimate (approved):
Let's go 🚀

PR #721, opened 2022-09-07
remove user admins when deleting org or deactivating user

axelsson:
When removing an org or user, we need to remove the user admin entry


axelsson:
[Remove user admin when deleting org or deactivating user](https://trello.com/c/75jd01QP/969-remove-user-admin-when-deleting-org-or-deactivating-user)

app/controllers/admin/organizations_controller.rb
```diff
@@ -45,6 +45,7 @@ def update
     end
 
     def destroy
+      @organization.user_admins.destroy_all
```
nanditagoclimate, line 48: What about the organization_id in users table?

axelsson: Good point, updated!

nanditagoclimate (approved):

PR #723, opened 2022-09-11
Add key to accounting reports

nanditagoclimate:
API response updated in [API doc](https://docs.google.com/document/d/1Yxu9UyQzFU_vUlBI1W-cWzvHPZCVw4oJ12BprC835-A/edit#)

Post deploy:
1. Run "heroku run rails db:migrate -a goclimateneutral"
2. Run the below code snippet: "heroku run rails c -a goclimateneutral"
Business::AccountingEmissionsCalculation.where(key: nil).each do |acc_em|
  if acc_em.other_emissions.present?
    acc_em.update!(other_emissions: JSON(acc_em.other_emissions), key: SecureRandom.hex(20)) 
  else
    acc_em.update!(key: SecureRandom.hex(20))
  end
end


kallenilver (approved):
Looks very good!

axelsson (commented):
As i understand this still doesn't solve the problem since we're using current_user.organization.id to fetch the report key. It is only a problem for us super admins when viewing other companies in CAT. We can still find the reports in admin so I don't see why this needs to be changed?

nanditagoclimate:
With this route with key as a param, even org_admins can download the report.

axelsson:
> With this route with key as a param, even org_admins can download the report.

Yes, IF the user is from the same org, which they could before this code was changed as well, using the id instead of the key. After these changes, that still works. It still doesn't work when viewing for example Enento while not being an org admin for them, and it doesn't work after these changes either. I've tested it locally.

nanditagoclimate:
You are right Johanna, this was discussed in the standup too on Friday. It is bad design to show our database ids in the URL. That is the whole point of using keys in both climate reports and in accounting climate reports.
Also, only the org_admin should be able to view their reports as this PR does. I've tested locally too before pushing the code.
Once the user flow is figured and we roll out CAT with org_admin support, this will fall into place too.

axelsson:
> 

Okay, I thought this whole PR was related to your comment in slack (https://goclimateneutral.slack.com/archives/C02DN5FRBDK/p1662642961399499) the other day, not just adding keys. A miss in communication. :)

axelsson (approved):

PR #727, opened 2022-09-15
allow super admins to download all acc reports

nanditagoclimate:
Super admins should be able to download all acc reports.


kallenilver (approved):
🥳

PR #730, opened 2022-09-15
Rubocop safe changes

kallenilver:
I noticed rubocop hasn't been run in a while, so I ran it in safe mode to get the code a bit more tight


nanditagoclimate (approved):
Good. 🏆

PR #737, opened 2022-09-26
Read only fortnor customers list

nanditagoclimate:
Since access_token are really long, they go on and on in the view. I tried a couple of things to word wrap it, but didn't succeed. So any pointers are welcome.


app/views/admin/fortnox_customers/index.html.erb
```diff
@@ -0,0 +1,33 @@
+<p id="notice"><%= notice %></p>
+
+<h1 class="heading-lg">Fortnox customers</h1>
+
+<table class="table">
```
kallenilver, line 5: I think a `style="word-break: break-all;"` on the `<table>` might work. Since I don't think we use tailwind in admin, otherwise this would work I think: https://tailwindcss.com/docs/word-break

kallenilver (approved):

app/views/admin/fortnox_customers/index.html.erb
nanditagoclimate: Great, thanks. I added this styling on the row itself otherwise all of the table got styled not so well.

PR #738, opened 2022-09-27
Automatically run migrations post deploy

nanditagoclimate:
This will help avoid cases where we manually forget to run migrations.


kallenilver (approved):
Good idea!

PR #740, opened 2022-09-27
Add company name in fortnox customers

nanditagoclimate:


app/controllers/fortnox_customers_controller.rb
```diff
@@ -32,6 +32,7 @@ def create
     fortnox_params = fortnox_api.connect(params[:code], params[:app])
 
     @fortnox_customer = FortnoxCustomer.new(
```
kallenilver, line 34: I think this will make the flow from fortnox crash since we are not sending in a company name there?

kallenilver (changes requested):

app/controllers/fortnox_customers_controller.rb
nanditagoclimate: Had a discussion this morning standup and it was decided that company name is not a must to have on the view for now so I shall close this PR.

PR #748, opened 2022-10-11
JSON bug fix

nanditagoclimate:
The older code worked when we made changes to the JSON data structure from the admin panel.
The changes in this PR also handles JSON updates we make through ruby console.


kallenilver (approved):
🚀

PR #753, opened 2022-11-09
Read monthly data

nanditagoclimate:
PR - 
1.  Get monthly data from CB and writes into AEC
2. Update sample data
3. Update test cases


nanditagoclimate:
Closing this as I'm opening one with both categories and monthly data support

PR #754, opened 2022-11-29
New categories+monthly data support

nanditagoclimate:


app/models/climate_calculator.rb
```diff
@@ -103,22 +120,26 @@ def emission_factors
...
   def climate_constants_for_sweden
     @climate_constants_for_sweden ||= YAML.load_file('config/country_wise_climate_constants.yml')['se']
   end
 
   def calculate_meal_emissions
-    return @emissions_data['Meals'] unless company_input['share_of_veg_meals']
+    return @category_groups['food_and_drinks']['restaurant'] unless company_input['share_of_veg_meals']
 
-    (@emissions_data['Meals'] * @company_input['share_of_veg_meals']/200) + (@emissions_data['Meals'] * (100 - @company_input['share_of_veg_meals'])/100)
+    (@category_groups['food_and_drinks']['restaurant'] * @company_input['share_of_veg_meals']/200) + (@category_groups['food_and_drinks']['restaurant'] * (100 - @company_input['share_of_veg_meals'])/100)
```
axelsson, line 142: I know it's not part of this PR but why is the first part here divided på 200?

axelsson: Yeah, either that or defined it as a variable with an explaining name. :)

nanditagoclimate: Shall do, thanks for pointing it out! 💯

kallenilver (approved):
I don't have the bandwith to do a thourough review, but feel free to deploy and see if it works!

PR #756, opened 2022-12-12
Handle no company input

nanditagoclimate:


app/controllers/api/v1/overviews_controller.rb
```diff
@@ -11,13 +11,17 @@ def show
         organization = Organization.find_by_key(params['key'])
         return organization_not_found unless organization
 
+        company_input = CompanyInput.where(organization_id: organization.id).last
+
+        return invalid_request_error unless company_input
```
nanditagoclimate, line 16: None of the other HTTP status codes made sense to me so I am sending 400 for companies that don't have a CompanyInput record. Does this make sense to you? @emilkaiser

emilkaiser: Ok!

emilkaiser (approved):

PR #760, opened 2022-12-15
add is_superadmin to user api response

axelsson:
Frontend needs to know about superadmin (for showing internal tools and stuff). Found out user api response crashed when user has no org so I added a fix for that.


emilkaiser (approved):
Nice!

I just (this minute) pushed a hack to the frontend for this. But they will work perfectly togethet!

nanditagoclimate:
Good work 🎖️

PR #771, opened 2023-01-24
Company data

axelsson:
* Added table for organizationData
* Added admin interface for that
* Changed in CC so that we only use org data instead of company input and so that it also fetches the right video for the right time period
* Added tests for that logic, to make sure we're fetching the relevant value for the right time period (example average_fte)
* Renamed/updated CC-tests

* There was a bug in CC regarding the electricity emissions so the numbers now and before are not entirely comparable. The bug was dividing the emissions in 12 (to make it apply to months rather than year) rather than dividing the consumption (app/helpers/climate_calculator_helper.rb in `calculate_electricity_consumption_emissions`).
* When replacing company_input with organization_data the current values in kwh needs to be divided in 12 to represent monthly values instead of yearly
* Previously we required one company_input with average_fte to display anything at all, now we require one org data with average_fte and we use the latest one regardless of end date. Might result in using old numbers, but that can be highlighted in CAT/pdf, to say new data is needed.
* Is the admin interface good enough for a first version?
* Do we need more tests?

I've compared the numbers in prod against local for Beyond retail, Olingo, PE Accounting and Win Win Ekonomi to make sure the numbers add up (which they almost do, except for the electricity numbers due to the bug)


app/helpers/climate_calculator_helper.rb
```diff
@@ -66,60 +66,74 @@ def scopes_average(scope_emissions_values)
     end
   end
 
-  def calculate_meal_emissions(monthly_emissions, company_input)
-    unless company_input['share_of_veg_meals'] && monthly_emissions['food_and_drinks']['restaurant']
+  def calculate_meal_emissions(organization_data, monthly_emissions)
+    share_of_veg_meals = organization_data[OrganizationDatum.keys[:share_of_vegetarian_meals]]
```
kallenilver, line 70: green feedback: It would be nice to be consistent in the naming of vegetarian vs veg

kallenilver: And I prefer vegetarian to differentiate it from vegan :)

app/helpers/climate_calculator_helper.rb
```diff
@@ -66,60 +66,74 @@ def scopes_average(scope_emissions_values)
...
 
     # Emissions of veg meal is considered as half of not veg meals
     veg_meals_emissions_compared_to_not_veg = 0.5
 
-    (monthly_emissions['food_and_drinks']['restaurant'] * company_input['share_of_veg_meals'] * veg_meals_emissions_compared_to_not_veg / 100) + (monthly_emissions['food_and_drinks']['restaurant'] * (100 - company_input['share_of_veg_meals']) / 100)
+    ((monthly_emissions['food_and_drinks']['restaurant'] * share_of_veg_meals / 100 * veg_meals_emissions_compared_to_not_veg) + (monthly_emissions['food_and_drinks']['restaurant'] * (1 - (share_of_veg_meals / 100)))).to_i
   end
 
-  def calculate_electricity_consumption_emissions(monthly_emissions, company_input) # rubocop:disable Metrics/MethodLength
+  def calculate_electricity_consumption_emissions(organization_data, monthly_emissions) # rubocop:disable Metrics/MethodLength
```
kallenilver, line 82: green feedback: I'm not sure the word `consumption` is needed, just to keep things shorter

db/schema.rb
```diff
@@ -307,6 +306,7 @@
   end
 
   create_table "fortnox_customers", force: :cascade do |t|
+    t.string "company_name"
```
kallenilver, line 309: 👀

kallenilver (approved):
I think this is good enough for now and safe enough to try 🚀

app/helpers/climate_calculator_helper.rb
axelsson: Niiice, good comments, I've made some updates to make it more consistent

db/schema.rb
axelsson: 🤷‍♀️ 🧠

axelsson:
[Organization data version/financial year handling](https://trello.com/c/iCrXUxIr/1288-organization-data-version-financial-year-handling)

PR #773, opened 2023-02-10
Remove crowdin and cleanup

axelsson:
1. Fixed the db seed to add superadmin + org admin roles, and org for demo with employee count so that demo can always be viewed in CAT directly after setup. Not sure why the old admin user was added with id 2, I changed it to 1 and added another example user as an admin for demo org.
2. Removed crowdin cause almost noone is using it, it's unnecessary overhead and I don't see a future where we make the website available in more languages than the ones we support now.

I've tested the removal of crowdin by visiting the local site in the different languages that we support but I'm not sure if this could affect us in any other way? As far as I've understood it there's no magic at build/deploy that needs to be changed after this.


kallenilver (approved):
Did not see this coming :) But yes, I agree, we havn't used crowedin in a year at least, so might as well remove it to keep things simple. 

I don't think it should effect us in any other way, should be safe to deploy as far as I know.

PR #774, opened 2023-02-14
Organization locations/manual emissions support

axelsson:
Added table `OrganizationEmissions` for adding manual emissions per financial year, which is connected to an organization. It contains emissions and scopes given as json, and a start and end date.

For `organization` `data_source` is added to know from where the data should be fetched. If set to `manual` we'll look for data in `OrganizationEmissions` instead of CB.

A separate `cb_key` has also been added to `organization`, in case we do not want to use the slug as the CB key to fetch the data. In this case this means we can create the organization Adtraction Sweden with the key `adtraction,adtraction_scandinavia,adtraction_group` which then will fetch combined data from these three organizations in CB. If the `cb_key` is not set, we'll use the regular key/slug. 

To test, add organization, set manual data, add organization emissions (emissions in this format: `{"office_space":3973,"business_travel":4947,"food_and_drinks":46,"electronics":1770}`, scopes in this: `{"2": 8.86, "3":4}`, add organization data -> employee count number, and then make a request to `http://localhost:3000/api/v1/overview?key=org_key`


db/schema.rb
```diff
@@ -195,23 +195,6 @@
     t.index ["organization_id"], name: "index_climate_reports_reports_on_organization_id"
   end
 
-  create_table "company_inputs", force: :cascade do |t|
```
kallenilver, line 198: Do we have a backup or screenshot of this data in case we want to replicate it for next calculations? Or is everything already migrated to the new format?

app/controllers/admin/organizations_controller.rb
```diff
@@ -21,22 +21,25 @@ def edit
     end
 
     def create
-      @organization = Organization.new(organization_params)
+      data_source = organization_params[:manual_data] == '1' ? 'manual' : nil
```
kallenilver, line 24: 🟢 It would be kind of nice to skip this conversion and send 'manual' from the form instead? If possible 🤷

app/controllers/admin/organizations_controller.rb
```diff
@@ -21,22 +21,25 @@ def edit
...
+                  notice: 'Organization was successfully created. Remember to add Employee count in Organization data.'
     end
 
     def update
       params[:users]&.reject(&:blank?)&.each do |user_email|
         @organization.add_admin_from_email(user_email)
       end
 
-      unless @organization.update(organization_params)
+      data_source = organization_params[:manual_data] == '1' ? 'manual' : nil
```
kallenilver, line 41: 🟢 same here

app/controllers/api/v1/overviews_controller.rb
```diff
@@ -5,26 +5,37 @@ module V1
...
       before_action :authenticate_user!
 
-      def show # rubocop:disable Metrics/PerceivedComplexity, Metrics/CyclomaticComplexity
-        return invalid_request_error unless params['key']
+      def show # rubocop:disable Metrics/PerceivedComplexity, Metrics/CyclomaticComplexity, Metrics/MethodLength
+        key = params['key']
+        return invalid_request_error unless key
 
-        organization = Organization.find_by(key: params['key'])
+        organization = Organization.find_by(key:)
```
kallenilver, line 12: nice

app/models/climate_bookkeeper_api_adapter.rb
```diff
@@ -1,33 +1,32 @@
...
 
-    make_request('financial-years')
+    make_request('financial-years', key)
   end
 
   private
 
-  def make_request(endpoint)
-    url = "https://bookkeeper-api.goclimate.com/api/v1/#{endpoint}/#{@key}"
+  def make_request(endpoint, key)
```
kallenilver, line 28: I like this small change: it's less magic, less prone to errors and easier to follow

db/schema.rb
kallenilver: Also, I don't see a migration for this?

kallenilver (approved):
💯 🔥

db/schema.rb
axelsson: My mistake, I've removed it locally on another branch, should not yet be removed.

app/controllers/admin/organizations_controller.rb
axelsson: Tried but failed, should be possible though, will put on TODO-list meanwhile :)

axelsson:
[Reintroduce locations support](https://trello.com/c/P4B8gzSJ/1299-reintroduce-locations-support)

PR #775, opened 2023-03-01
Dont send payment failed mails to users that has not paid in the last…

kallenilver:
… six months

I'm worried that we send lots of payment failed-mails to users who hasen't paid in quite a while and therefor might risk getting spam-classified.


axelsson (approved):
Looks like a good way to decrease the number of mails. Maybe we can talk a bit about why this happens at all, shouldn't we be able to inactivate users with too many failed payments?

PR #778, opened 2023-03-14
Upgrade packages

axelsson:
* Ruby verison is updated, 3.1.2 -> 3.2.1
* Updated orbs & images in circleci
* Updated webpack and surrounding packages to avoid security problems, which also was needed to step up the node version to 18 in heroku.
* Updated some gems that was needed/removed unused one
* Renamed CSS component `collapse` that was named the same as a keyword in tailwind, they clashed after webpack update

To test this, install node v18 and ruby 3.2.1, run `bundle install` and `yarn install`

I've tested all I can think of on the webpage (signup flow, pdf downloads, public/private api's). Not tested: fortnox things, not sure how.

Not sure if any more changes are required for deploying to Heroku, we'll see.


kallenilver (approved):
Looks good! Nice to be up to date! 🚀

PR #782, opened 2023-04-12
add supported projects to public emissions API, add org id to invoice…

axelsson:
… table


app/controllers/api/v1/public_emissions_controller.rb
```diff
@@ -27,10 +27,28 @@ def show
...
+          supported_projects: supported_projects(organization.id)
         }
       end
 
+      SupportedProject = Struct.new(:calculation_period, :name, :co2_offset)
+
+      def supported_projects(organization_id)
+        climate_report_projects = ClimateReport.where(organization_id:).joins(:invoice)
+                                               .map do |c|
+          next if c.invoice.project&.name.blank?
```
kallenilver, line 39: I guess it would be a bit nicer to skip this row and have a .joins(:project) instead if that works?

app/controllers/api/v1/public_emissions_controller.rb
```diff
@@ -27,10 +27,28 @@ def show
...
+
+          SupportedProject.new(c.calculation_period.to_i, c.invoice.project.name, c.invoice.co2e.co2e)
+        end.compact
+
+        invoice_projects = Invoice.where(organization_id:).joins(:project).map do |i|
+          next if i.project.nil?
```
kallenilver, line 45: is this row really needed? I thought joins is a inner join that would exclude nil projects?

kallenilver (approved):
Looks good!

app/controllers/api/v1/public_emissions_controller.rb
axelsson: Done! Had problems with this one but that was due to super broken local test data.

app/controllers/api/v1/public_emissions_controller.rb
axelsson: True!

PR #788, opened 2023-04-26
Create database table and model

cnyg:
We start with the database table to keep the pull request small for ya


axelsson (approved):
Fint! Snackade vi inte om att spara ett userId också?

PR #791, opened 2023-05-05
Remove cloud servers green share, add more flexible org data entry instead

axelsson:
* Removed cloud servers green as a subcategory entirely
* Removed org data key "green_cloud_servers_share"
* Added "cloud_servers" instead, taking +- kg co2e/mon
* CC now checks for org data entries with the same names as subcategories, and in that case adds the value
* Changed order of things in CC to apply org data on a flat structure (structuring into categories now happens later)
* Had to change the enum for org data keys into a hash instead, to ease showing the description in admin
* This has a shitty drawback of us now mentioning keys as simple strings instead of for example `OrganizationDatum.keys[:revenue_msek]` so it'll be harder to keep track of where they're used
* Added description to org data, included in API answer
* Set default start date of org datum to first day last year

Let's start with only "cloud_servers" for now and we'll see what the next thing to add is. I've added support for showing the description of an org datum in "assumptions" in CAT & PDF already. It's limited to only language for now (since it's only one extra field...) but if it's needed I guess we could add "description_sv" and "description_en" or something like that. Most reports are delivered in one language.


axelsson:
[Change cloud servers green handling](https://trello.com/c/VgXWiQmA/1551-change-cloud-servers-green-handling)

app/helpers/climate_calculator_helper.rb
```diff
@@ -5,19 +5,16 @@ module ClimateCalculatorHelper
   def override_with_organization_data(organization_data, monthly_emissions)
     return monthly_emissions if organization_data.nil?
 
-    if monthly_emissions['food_and_drinks']
-      monthly_emissions['food_and_drinks']['restaurant'] = calculate_meal_emissions(organization_data, monthly_emissions)
-    end
+    monthly_emissions['restaurant'] = calculate_meal_emissions(organization_data, monthly_emissions) if monthly_emissions['restaurant']
```
frusunnanbo, line 8: Gud va najs att slippa massa if-satser :star2:

app/helpers/climate_calculator_helper.rb
```diff
@@ -5,19 +5,16 @@ module ClimateCalculatorHelper
...
+    monthly_emissions['electricity'] = calculate_electricity_emissions(organization_data, monthly_emissions)
+    monthly_emissions['heating'] = calculate_heating_emissions(organization_data)
+    monthly_emissions['home_office'] = calculate_work_from_home_emissions(organization_data)
 
-    if monthly_emissions['other_purchases']['cloud_servers']
-      monthly_emissions['other_purchases']['cloud_servers'], monthly_emissions['other_purchases']['cloud_servers_green'] =
-        calculate_cloud_servers_emissions(organization_data, monthly_emissions)
-    end
+    organization_data.each do |key, value|
+      next unless monthly_emissions.key?(key) & !value.nil?
```
frusunnanbo, line 14: A couple of questions / discussion points: 
 * Do we really want to specify a monthly value (Maybe this is already discussed and clear, we don't know :-))
 * If we have specified e.g. a monthly cloud_servers org data of -20 for one year, based on a manual calculation o f yearly emissions spread out over 12 months and we have no cloud server emission in our accounting data e.g. for February, then the total will be wrong...
 
 Happy to talk about this next week!

frusunnanbo (commented):
Looks great! :fire: :1st_place_medal: 

We have a couple of questions, see below.

frusunnanbo:
What happens for customers that already have the old cloud_servers_green_share organization datum? Will their calculations be right?

config/climate_bookkeeper/sub_category_to_category_mapping.yml
```diff
@@ -6,6 +6,7 @@ waste: office_space
 electricity: office_space
 cooling: office_space
 heating: office_space
+home_office: office_space
```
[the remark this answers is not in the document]
axelsson: Den fanns inte 😬 Skapades "manuellt" i CC-helper efter mappningen som skedde tidigare förut.

axelsson:
> What happens for customers that already have the old cloud_servers_green_share organization datum? Will their calculations be right?

No one is using it atm so nothing to worry about.

app/helpers/climate_calculator_helper.rb
axelsson: * Since we're always applying the org data on month basis we'd otherwise have to let the input be per year and apply it/12 instead (which is doable).
* You could in that -20-case add several inputs with shorter dates and a value that will still add upp to -20 and "skip" the month with no cloud server emissions. The current solution supports adding -20 for just one month if that's the dates you set on it and if that makes sense. If you look into admin https://www.goclimate.com/admin/organization_data you can see how electricity_kwh has been applied in a few cases, for example.

PR #808, opened 2023-06-21
if wfh share is missing, assume 100%, if wfh count is missing, assume…

axelsson:
… all employees

Today you gotta add two org data fields for work from home emissions. In a lot of cases they're either fully remote (100%) or all of the employees are working from home at some share. To avoid having to input both fields at all times, assume that the share targets all their employees or that the number wfh are 100% remote if the other piece of info is missing. Will not affect current customers data since they're all using both fields at the same time today.


cnyg (approved):
Good change! :)

PR #816, opened 2023-08-18
Google tag manager update

kallenilver:
Inserting new google tag manager blocks
- GTM_TRACKING_ID is needed in .env and/or prod

This removes GA4 tracking so that needs to be readded, but in GTM instead

Also:
- we need to add GTM to the blog
- we need to add it to the new site
- we should also remove tracking to facebook and linkedin, and re-add to GTM if needed


axelsson (approved):
Schysst, verkar funka (testade https://tagassistant.google.com/ på localhost och den dök upp)

PR #817, opened 2023-09-11
New navigation

axelsson:
All changes I could think of for the migration to the new landing page/use next app instead.
 
* Update buttons styling to better match new design
* Let / lead nowhere (will be routed through nginx to next app), add /individuals instead
* Update sitemap with /individuals along with some new urls to the app
* Add separate header and footer for business related pages
* Change header & footer in affected pages
* Allow some pages to have any of them depending on url param "individuals=true"
* Let links to the new app use 'sv' as lang instead of 'se'
* Use minimal header for error pages to avoid confusion
* Remove landscape-footer for all business related pages
* Remove old "announcement" on individuals landing page
* Update all links I could find that should have ?individuals=true in url

Thoughts
* Should we change the url-param to something shorter/more discrete?
* Is there anything more that could break due to these changes? 😅
* I guess we'll need to update links to our individual page in SoMe and things like that
* The blog still holds a really old menu, it should probably be updated soon as well
* Should we (somewhere) add a redirect from /business to / when going live? (There are several redirects to business in routes.rb today)


kallenilver (approved):
Looks good! 🔥 

- I think the url-param works
- a redirect from /business doesn't hurt
- yes, blog and some needs an update after this
- can't think of anything that will break :)

axelsson:
Updated with
* Redirects /business to /
* remove old cookie banner
* link 'get started' in business menu to services page

PR #822, opened 2023-10-10
add endpoint for deleting oxceed connection

axelsson:
For oxceed to be used when a company's contract has run out and they've not renewed it.

* delete endpoint for oxceed added
* todo left for when we've decided what we want and need to do on a delete
* sends mail to us letting us know which company was canceled. Since this is gonna be triggered yearly from oxceeds system, I assumed there's no user connected to the action.


frusunnanbo (approved):
:star2:

PR #825, opened 2023-10-23
Use category entity

axelsson:
I've added category as entity in the db
It has the following properties
* key (to match the current ones)
* name_en, name_sv, names to be sent in api answer (to oxceed among others) and in the future to be used in CAT as well
* scope, a number today, might need updating if we support subscopes in the future but will do for now
* emission_factor_key, wanted it to be a foreign key to the emission factor table but since there can be multiple instance of each emission factor it's a string. When creating the category, only existing emission factor keys are given as options
* reference to another category, its main category to keep track of the hierarchy

I've added the main categories in the seeds file, and a rake task `add_categories` to populate the db with all the existing categories. It will use the current yml-files to set the correct main category, scope and emission factor. I guess this should be removed once we've created the correct data in prod, and then we should use prod db backup data to get a decent working env instead.

Right now emissions within categories that can't be found won't be included in calculations. As before, an email will be sent if we fint missing categories.

The names of the categories are included in the API response, ex 
```						
"key": "other_purchases",
"name_en": "Other purchases",
"name_sv": "Övriga inköp",
"value": 5777,
```


kallenilver (approved):
Snyggt! Jag har lite svårt att överblicka climate_calculator och emissions_helper, men kan inte se några uppenbara fel. Allt annat ser bra ut tycker jag!

spec/requests/api/v1/partner/oxceed/organizations_controller_spec.rb
frusunnanbo, line 1: This does not seem to belong to categories?

app/controllers/api/v1/partner/oxceed/organizations_controller.rb
frusunnanbo, line 1: Is this related to the categories change?

app/helpers/emissions_helper.rb
```diff
@@ -45,25 +45,28 @@ def format_period_data(monthly_data, financial_years, organization_data, include
...
+        sub_categories = all_sub_categories_keys.map do |sub_category_key|
+          sub_category_total = all_sub_categories.select { |s| s[:key] == sub_category_key }.map(&:value).sum
           # All green electricity goes into scope 3
-          if sub_category == 'electricity' && !green_electricity_share.nil?
+          if sub_category_key == 'electricity' && !green_electricity_share.nil?
             scopes[2] += sub_category_total * (1 - (green_electricity_share / 100))
             scopes[3] += sub_category_total * green_electricity_share / 100
-          else
-            scopes[sub_category_scope_mapping[sub_category]] += sub_category_total
+          elsif all_categories[sub_category_key]&.scope
```
frusunnanbo, line 59: When would scope not exist?

And in that case, would we miss something in the scope total?

app/helpers/emissions_helper.rb
```diff
@@ -78,21 +81,23 @@ def format_period_data(monthly_data, financial_years, organization_data, include
     end
   end
 
-  def format_monthly_data(emissions_data)
+  def format_monthly_data(emissions_data) # rubocop:disable Metrics/MethodLength
```
frusunnanbo, line 84: Could chatGPT help us untangle this code a bit?

db/migrate/20231011054313_create_categories.rb
```diff
@@ -0,0 +1,13 @@
+class CreateCategories < ActiveRecord::Migration[7.0]
+  def change
+    create_table(:categories) do |t|
+      t.string(:key, null: false)
+      t.string(:name_en, null: false)
+      t.string(:name_sv, null: false)
+      t.integer(:scope)
```
frusunnanbo, line 7: When could this be null?

frusunnanbo: Ah! When it's main category!

frusunnanbo (approved):
Looks great! :star2:

A few questions / comments inline

spec/requests/api/v1/partner/oxceed/organizations_controller_spec.rb
axelsson: True, the tests gave warnings which i fixed  by these changes 😬

app/controllers/api/v1/partner/oxceed/organizations_controller.rb
axelsson: nope!

app/helpers/emissions_helper.rb
axelsson: I wish, men den hade inte så mycket bättre förslag tycker jag.

app/helpers/emissions_helper.rb
axelsson: Good finding! It's not needed at all any more since subcategories that doesn't exist already have been filtered out earlier. I've updated it to a simple else.

axelsson: And missing subcategories works as previously: they're excluded and you get an email about them

PR #826, opened 2023-10-26
Update employees from oxceed

axelsson:
Added a rake task to update organization data employee count for oxceed companies with the data from CB (or Oxceed actually). Added a field ` source` in Org data table to keep track of where changes come from, it's intentionally read only in admin right now.

The job only creates the necessary org data entries, if two months in a row holds the same number, no new entry will be created until it's changed, like the way we generally use org data today.

I added a check on whether there are new entries to update with before deleting them to ensure we're never entirely without a number for employees for a company. I also wrapped that part in a transaction for that purpose.

The task itself runs via a job that is to be added in Heroku (scheduled jobs can be seen here: https://dashboard.heroku.com/apps/goclimateneutral/scheduler ) so that part is not in the code.


lib/tasks/update_oxceed_employee_count.rake
```diff
@@ -0,0 +1,42 @@
+desc 'Update employee count for Oxceed subscriptions'
+task update_employee_count_for_oxceed_subscriptions: :environment do # rubocop:disable Metrics/BlockLength
+  def values_differ(current, previous)
```
kallenilver, line 3: A thing of taste but I would not do a method out of this

kallenilver (approved):
Looks good! 🚀

lib/tasks/update_oxceed_employee_count.rake
axelsson: agreed, i think there was more logic in this when i began, i've removed it, seems simpler :)

